PR #8, opened 2019-08-17
Fix wasLastUpdatedBefore to be in days, not hours

jrbriggs:
The parameters and variables names make it seem like this is the intended behavior.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -92,7 +92,7 @@ function isLabeledStale(issue: Issue, label: string): boolean {
 }
 
 function wasLastUpdatedBefore(issue: Issue, num_days: number): boolean {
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -92,7 +92,7 @@ function isLabeledStale(issue: Issue, label: string): boolean {
 }
 
 function wasLastUpdatedBefore(issue: Issue, num_days: number): boolean {
-  const daysInMillis = 1000 * 60 * 60 * num_days;
+  const daysInMillis = 1000 * 60 * 60 * 24 * num_days;
```
localheinz: Perhaps it makes sense to adjust to

```suggestion
  const daysInMillis = num_days * 24 * 60 * 60 * 1000;
```

?

PR #9, opened 2019-08-20
Fix typo

vincentschen:
iusses >> issues


vincentschen:
Nevermind, looks like this was fixed in master.

PR #62, opened 2020-05-08
Bumping actions/github to 2.2.0

pjquirk:
This adds support for running against GHES by using [v2.2.0](https://github.com/actions/toolkit/blob/master/packages/github/RELEASES.md#220) of `@actions/github`.

Tested by pushing this version of the action to an "actions" org on the GHES instance.  I also needed to set `NODE_TLS_REJECT_UNAUTHORIZED = 0` temporarily in the action to get around my self-signed (or otherwise not trusted) server cert.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -43,6 +43,56 @@ module.exports =
 /************************************************************************/
 /******/ ({
 
+/***/ 0:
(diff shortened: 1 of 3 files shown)


hross:
@pjquirk I'm updating this action so I'll do this another PR to avoid conflicts.

PR #66, opened 2020-05-12
Fix for logging error types instead of strings in the main try/catch

pjquirk:
When an `Error` type is thrown instead of a `string`, the command  `core.error(error)` itself would fail, which would mask the real error.  This generally results in an error that is described in https://github.com/actions/toolkit/issues/398.

This bug has been fixed in 1.2.4 of `@actions/core`, so let's just use that version.

### DIFF
--- a/src/IssueProcessor.ts
+++ b/src/IssueProcessor.ts
@@ -374,6 +374,11 @@ export class IssueProcessor {
       event => event.event === 'labeled' && event.label.name === label
     );
 
+    if (!staleLabeledEvent) {
+      core.warning(`Could not find when issue #${issue.number} was labeled with ${label}`);
+      return '';
+    }
+
     return staleLabeledEvent!.created_at;
   }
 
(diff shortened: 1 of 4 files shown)


pjquirk:
The failure in `test` [doesn't appear to be new](https://github.com/actions/stale/runs/663767410#step:3:14), this PR just makes it be correctly reported as an error now 😄 / 😭

src/IssueProcessor.ts
```diff
@@ -374,7 +376,12 @@ export class IssueProcessor {
       event => event.event === 'labeled' && event.label.name === label
     );
 
-    return staleLabeledEvent!.created_at;
+    if (!staleLabeledEvent) {
```
pjquirk, line 379: @hross FYI

hross: awesome. thanks for fixing my bad code :)

PR #68, opened 2020-05-13
add option to remove flags when updated

dev-zero:


dev-zero:
a successful test run can be found here: https://github.com/dev-zero/stale-issue-test/issues?q=is%3Aissue

dev-zero:
if needed I can also split-up the variable for separate PR and Issue labels

dev-zero:
Ok, I just realized that even with this feature I'm missing a piece. So here is the the use case:

* we want to mark issues which require additional input from the reporter with `needinfo`
* only for issues labeled with `needinfo` we want to run the stale bot
* as soon as anyone except the stale bot adds a comment should the labels `needinfo` and `stale` be removed (if present)

With this PR the stale bot can be taught to remove `needinfo` together with `stale` IF the issue has already the `stale` label, but not if it only has the `needinfo` label because for that must the stale bot also check the `needinfo` label creation date against the last activity on the issue.

Any ideas?

Putting the management of the `needinfo` label in a separate action makes matters more complicated because then I'd have to explicitly filter a comment added by the stale bot.

dev-zero:
would you be ok if for labels in `remove-labels-when-updated` is executed independently of the `stale` label? Meaning that for each label in `remove-labels-when-updated` it checks the labelCreationDate against the updated_at of the issue and if updated_at is newer it will remove the label?

dev-zero:
> would you be ok if for labels in `remove-labels-when-updated` is executed independently of the `stale` label? Meaning that for each label in `remove-labels-when-updated` it checks the labelCreationDate against the updated_at of the issue and if updated_at is newer it will remove the label?

This is what this PR now does

hross:
I'm a bit swamped and will look at this next week.

hross:
I am not sure about this change. I understand the desire to simplify the workflow and not have another action, but I don't want to add unnecessary complexity to the action itself with a bunch of options for cases that can be handled with custom workflows and other actions.

Right now I am going to close this as I don't think it fits the spirit and you can do this same thing by stealing code from this action and checking the commenter. "Need info" or other labels are really other workflows independent of stale.

Issue #79, opened 2020-05-21
[Question] operations-per-run parameter?


doctorperceptron:
What exactly is the `operations-per-run` parameter? I tried running this action in my repo in debug mode and got a warning "Reached max number of operations to process. Exiting." without any other output.

aandergr:
I also wondered about the warning "Reached max number of operations to process. Exiting." after updating from v1 to v3 and doing some other changes in my configuration. 

`operations-per-run` is meant to be "The maximum number of operations per run, used to control rate limiting." During the action's run, the number of operations is counted, where with having a brief look at the code, the following operations seems to count:
- Accessing a page of issues is **one operation** (issues – filtered by state 'open' and onlyLabels – are processed in pages of 100 issues each)
- Closing a stale issue is **one operation**
- Removing a stale label is **one operation**
- Processing an issue that is stale is **one or two operations**, one each for obtaining its date of being marked stale and for checking whether it is still stale 
- Marking an issue stale is **four operations**: Two are counted in https://github.com/actions/stale/blob/96b682d29f2b26168a8b283c62da1ad3f1262120/src/IssueProcessor.ts#L172 plus two in https://github.com/actions/stale/blob/96b682d29f2b26168a8b283c62da1ad3f1262120/src/IssueProcessor.ts#L303 I guess this should be 2 (create comment + add label), rather than 4.

Further, it seems that only after processing a complete page of issues it is checked whether there are still operations left, so the first 100 issues seem to be always processed regardless the `operations-per-run` parameter, making it hard to limit the stale action to only process a few issues at a time.

doctorperceptron:
Thanks @aandergr. What's the purpose of limiting the number of operations? Can we turn this off, to guarantee all the issues are processed?

aandergr:
[...]

PR #81, opened 2020-05-25
Fix minor issues regarding counting of operations

aandergr:
This addresses two minor issues regarding how operations are counted, regarding the `operations-per-run` parameter:

- Count markStale() as 2 operations rather than 4.
- Evaluate after each processed issue whether there are operations left, rather than after each page of issues.

I commented these observations in https://github.com/actions/stale/issues/79#issuecomment-633510698.


hross:
Fixed this in https://github.com/actions/stale/pull/83 instead (made some additional changes)

PR #87, opened 2020-05-28
Add optional Close Message

Chocrates:
👋  Hi All,

We needed the ability to add a comment to the issues once the stale bot closed them so our customers had a better idea of why it was closed and what the next steps they needed to take to action their item.  

Here is what we added:

* Added `close-pr-message` and `close-issue-message` to get closing message
* Changed the close logic so that ancient issues won't immediately close if they are past the both the stale and close date
* Added a mocking library so we can test the octokit calls without needing to actually talk to the api
* Think we ran eslint and prettier because I see a lot of formatting changes 😐 

I am happy to talk about if you'd like any or all of this.  I'll just have to figure out how to break apart the pieces youd like if so 😄 

cc: @froi @zkoppert @jasonmacgowan

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug Jest Tests",
+            "type": "node",
+            "request": "launch",
+            "program": "${workspaceRoot}/node_modules/jest/bin/jest",
+            "args": [
+                "-i"
+            ],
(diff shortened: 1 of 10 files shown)


Chocrates:
Fixing the merge conflicts here: https://github.com/ActionsDesk/stale/pull/7
Once that gets merged in this should update (I hope).

Chocrates:
Closing for now, appears to be a bug in this.  I'll re-open once the bug is fixed and master is merged in.

PR #93, opened 2020-06-08
Fix stale closing logic by removing close date addition

hross:
The closing logic doesn't work because label updates push the updated date on an issue.

I originally added closing days + stale days thinking the `updated` time wouldn't take the label add into account and because there was a bug where we didn't update the issue `updated_at` in memory when we marked it stale.

Although there were tests, without the API results its hard to specifically mock all the API edge cases. Existing tests should cover the rest of this fix to make sure there are no regressions.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -43,6 +43,56 @@ module.exports =
 /************************************************************************/
 /******/ ({
 
+/***/ 0:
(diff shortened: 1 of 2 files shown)


hross:
Going to wait to see if I can get some live tests on repos that have stale issues which should be closed before I merge and publish a new version.

PR #103, opened 2020-07-08
Allow for issues/PRs to be closed after manually being marked as stale

phillipj:
Previously this action would stop and skip all issues or pull requests it found if no `stale-issue-message | stale-pr-message` option had been configured.

In practise that meant this action could not be used in repositories where the maintainers would want to manually mark issues and pull requests as stale, instead of this action being responsible for the said marking.

That's what these changes allows for, maintainers being responsible for marking, but still want help from this action to close those issues or pull requests X days after they got manually marked as stale by labelling them.

Configuring `days-before-stale: -1` will activate that behaviour, inspired by how `days-before-close: -1` works already.

Would that sound like a good fit for this action by any chance?

### DIFF
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -107,6 +107,31 @@ test('processing an issue with no label will make it stale and not close it if d
   expect(processor.closedIssues.length).toEqual(0);
 });
 
+test('processing an issue with no label will not make it stale if days-before-stale is set to -1', async () => {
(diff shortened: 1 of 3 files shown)


phillipj:
Forgot to cross link this with https://github.com/nodejs/github-bot/issues/261 where we've discussed using this action to close issues and pull requests in [nodejs/node](https://github.com/nodejs/node).

Issue #78, opened 2020-05-21
Add an option to skip adding a comment while marking issue / pr as stale


mollyIV:
### Description 

Currently there's no option for **not adding** a comment while marking an issue or a pull request as stale. As a real life example use case, a bot on [`fastlane`](https://github.com/fastlane/fastlane) repository only adds a label [`status: needs-attention`](https://github.com/fastlane/fastlane/labels/status%3A%20needs-attention) to the pull requests that have been hanging without any action for a while. It's a handy way to express the need for an action to be taken without an explicitly pinging a given person. 

### Proposal

I'd like to propose adding two parameters to the action's inteface:

```yaml
skip-stale-pr-message:
  description: 'Skip adding stale message when marking a pull request as stale.'
  default: false

skip-stale-issue-message:
  description: 'Skip adding stale message when marking an issue as stale.'
  default: false
```

No breaking changes, only opt-in. 

Empty `stale-pr-message` / `stale-issue-message` will always take a precedence over  `skip-stale-pr-message` / `skip-stale-issue-message`.

### Workaround 

Create a custom GitHub Action based on an official `stale` action implementation. 

I'll gladly implement this feature if a proposal is accepted 😊

pantharshit00:
I made a PR for this: https://github.com/actions/stale/pull/119

PR #119, opened 2020-07-22
feat: add skip-stale-pr-message and skip-stale-issue-message

pantharshit00:
Fixes https://github.com/actions/stale/issues/78
This introduces two new flags `skip-stale-issue-message` and `skip-stale-pr-message` which would skip the message addition if these are set to true.

They are backward compatible as empty `stale-issue-message` or `stale-pr-message` will take precedence over them.

Regarding the tests, I had to mock a private method in order to test this. Ideally I would mock the methods on the github client but all tests seem to be running in debug mode so I didn't do that. If you have a better proposal for adding tests for this, I would be happy to change this.


pantharshit00:
Example usage: https://github.com/prisma/prisma/blob/master/.github/workflows/label-stale-issues.yml

PR #175, opened 2020-09-19
Delete licemce

Piretehran:


kadler:
Why would the license be removed? This looks like a spam PR for Hacktoberfest.

PR #184, opened 2020-10-02
Commit changes to dist/index.js from latest dependency upgrades

rmacklin:
It seems that several of the recent dependency upgrades courtesy of dependabot (e.g. #155 and #163) have caused the committed `dist/index.js` file to become out of sync with the output of a fresh compilation. This PR updates `dist/index.js` with the output from running a fresh `npm install` and `npm run all`.

This should make it so that future changes don't include large, unrelated diffs in `dist/index.js`.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1,63 +1,917 @@
 module.exports =
-/******/ (function(modules, runtime) { // webpackBootstrap
-/******/ 	"use strict";
-/******/ 	// The module cache
-/******/ 	var installedModules = {};
-/******/
-/******/ 	// The require function
-/******/ 	function __webpack_require__(moduleId) {
-/******/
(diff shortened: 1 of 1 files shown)


hross:
@rmacklin thanks for this PR! I apologize I am going to close it and link to the same thing but for one I generated locally. Just doing that so I can review/check it myself rather than try to review the diff and take a security risk on the distribution. Really appreciate it though!

rmacklin:
Ok, thanks! For posterity, the changes were identical (I'm glad to know our machines produced the same diff):
https://github.com/actions/stale/compare/0bec8142598bbeee4c22a27e448494f06e0aa928..44f9eae0adddf72dbf3eedfacc999f70afcec1a8

Issue #21, opened 2019-09-13
New comments do not remove the Stale label


gergelyke:
As raised in https://github.com/uber-web/baseui/pull/1892, it seems like new comments do not remove the Stale label. Do we miss anything in our configuration?

```yml
name: 'Close stale issues'
on:
  schedule:
    - cron: '0 0 * * *'

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v1
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          stale-issue-message: "This issue is stale because it has been open 30 days with no activity. If it's still valid, please remove the stale label or comment on the issue, otherwise this ticket will be closed in 5 days"
          days-before-stale: 30
          days-before-close: 5
```

LekoArts:
I'd also be interested in knowing if this is a bug or user error.

eileenmcnaughton:
I am +1 on that feature being useful!

MikeyMJCO:
Also +1 on the feature being useful :-)

LekoArts:
@gergelyke @eileenmcnaughton @MikeyMJCO 
I've implemented this feature in our fork here: https://github.com/gatsbyjs/stale/blob/128ed907d5174a32717e7761e63294f5c747d8e2/src/main.ts#L122-L130

It lists all comments since the "days-before-stale" and filters for bot/user comments. If it can find user comments, the resulting array is not empty and hence the function return true.

ruudk:
@LekoArts so cool, did you consider proposing this as a PR to this repo? Would love this to be standard feature ❤️

eileenmcnaughton:
Agree - this needs to be a standard feature before the main feature is usable

LekoArts:
I changed the logic in the meantime to a more robust way of checking it: https://github.com/gatsbyjs/stale/blob/gatsby-refactoring/src/main.ts#L143

This is all WIP so use it at your own risk, we're still testing ourselves. And if they maintain the repository again and would accept a PR we gladly would PR that upstream.

eileenmcnaughton:
Hmm so the upstream repo is not maintained anymore :-(

TheSalesOpsGuy:
Were you ever able to merge your fork/updates into the main action? I'm trying to get the old probot functionality back as well (remove label if commented).

LekoArts:
You can use our fork if you want: https://github.com/gatsbyjs/stale
[...]

PR #192, opened 2020-10-09
Stop filtering comments by `github.context.actor` when determining if an issue is still stale

rmacklin:
As originally mentioned in https://github.com/actions/stale/pull/73#discussion_r498578074, I added a scheduled workflow with the stale action and it seems that when the action runs, `github.context.actor` is myself, not the github actions bot. As a result, an issue that I had commented on after it was labeled `stale` ended up getting closed anyway. Here's what's in the logs for the action (incorrectly claiming that I am a bot):
```
Checking for comments on issue #15419 since 2020-09-22T00:06:03Z
Comments not made by rmacklin or another bot: 0
```

It seems like others have run into the same issue: https://github.com/actions/stale/issues/21#issuecomment-656002085.

According to this post in the GitHub Support Community, it seems to be expected behavior that the actor will be the user who set up the schedule:
https://github.community/t/who-will-be-the-github-actor-when-a-workflow-runs-on-a-schedule/17369

Thus, it doesn't seem appropriate to ignore comments from `github.context.actor` when determining if there has been activity since the issue was labeled stale; that is not a bot user as previously thought when this behavior was originally added in 96b682d29f2b26168a8b283c62da1ad3f1262120. So in this PR I'm proposing we stop filtering out those comments.

### DIFF
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -618,8 +618,8 @@ test('stale label should be removed if a comment was added to a stale issue', as
   expect(processor.removedLabelIssues.length).toEqual(1);
 });
 
-test('stale label should not be removed if a comment was added by the bot (and the issue should be closed)', async () => {
-  github.context.actor = 'abot';
+test('stale label should be removed if a comment was added by the github actor (and the issue should not be closed)', async () => {
+  github.context.actor = 'person-who-scheduled-stale-action';
   const TestIssueList: Issue[] = [
     generateIssue(
       1,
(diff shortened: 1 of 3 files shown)


rmacklin:
@hross Do you have any thoughts on this?

hross:
Unfortunately this is a limitation of the scheduling mechanism at the moment. The user who set up the schedule will be the "bot". In most cases this is desired, since the bot user will add issue labels and other comments on *stale* and we don't want those to count, even if it introduces the problem of excluding the created user.

If you remove the check the stale bot won't be able to close issues after marking them stale because they will then count as "not stale" due to the bot's actions.

PR #199, opened 2020-10-19
fix(label): allow to use spaces inside the labels

### LINKED ISSUE #98: Multiple exempt-pr-label where labels have spaces
I would like to have two labels to exclude this action. From the examples I could find, a comma separated quoted string appears to be the correct way. However, most of our labels have spaces, unlike the examples. The following represents our configuration, however, it still triggered the action and closed a PR with *both* labels.

- Did I miss some additional configuration?
- Could a better example be provided?

```
---
name: "Close Stale Pull Requests"
on:
  schedule:
  - cron: "0 0 * * *"

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/stale@v3
      with:
        repo-token: ${{ secrets.GITHUB_TOKEN }}
        stale-pr-label: 'stale'
        exempt-pr-label: 'work in progress,Do Not Merge'
```

C0ZEN:
@hross since the tests were not handling the case on error I am not sure that this changes really fix the problem but I strongly think that this was related to the locale string stuff.
Using instead deburr from LoDash seems a good thing but once again I count on you to tell me if you also think that this is right.

Fixes #98.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 node_modules/
 lib/
 __tests__/runner/*
+.idea
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,10 +2,10 @@ module.exports = {
   clearMocks: true,
   moduleFileExtensions: ['js', 'ts'],
   testEnvironment: 'node',
-  testMatch: ['**/*.test.ts'],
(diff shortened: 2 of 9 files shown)


hross (approved):

16987814:
my kids are not for sale. atte spider

PR #212, opened 2020-11-16
Added date-field option

artemsaveliev:
We are using stale, but update labels based on when issue is created - not when it's last updated.
This is because while there may be discussion on the issue, we still want to force the staleness indicator - primarily to drive certain SLAs.


hross:
I don't think we will add this for now due to complexity of options.

PR #220, opened 2020-11-22
fix(remove-label): do not encode the label to make sure to remove it

### LINKED ISSUE #173: 'Label does not exist' error when removing issues from stale state
Apparently starting from version v3.0.9 (based on the date it started to happen) I started to get the following error when the actions was removing the stale label of an issue:
```
##[error]Error removing a label: Label does not exist
```

This was the first run it happened: https://github.com/RocketChat/Rocket.Chat/runs/907729444?check_suite_focus=true#step:3:401

This is letting all issues still marked as stable, not sure it is also preventing them to be closed.

As [you can see](https://github.com/RocketChat/Rocket.Chat/actions/runs/181500400/workflow#L37) we use the `stat: stale` label, which is [still valid](https://github.com/RocketChat/Rocket.Chat/labels/stat%3A%20stale)

It took me a long time to notice this, as we have too many issues 🙈

C0ZEN:
Remove the `encodeURIComponent` when calling the GitHub API to remove a label.
The documentation is not specifying anything regarding special comments so it could work but I was not able to test it.

Fixes #173.

### DIFF
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -195,8 +195,13 @@ test('processing a stale issue containing a space in the label will close it', a
     )
   ];
 
+  const opts: IssueProcessorOptions = {
+    ...DefaultProcessorOptions,
+    staleIssueLabel: 'state: stale'
+  };
+
   const processor = new IssueProcessor(
-    DefaultProcessorOptions,
+    opts,
     async p => (p == 1 ? TestIssueList : []),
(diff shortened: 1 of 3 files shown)


jvalkeal:
Tried this pr and it works ok with slashes with label. Could maintainers check these changes!

hross (approved):

C0ZEN:
Thank you

PR #256, opened 2020-12-16
Update action.yml

Heretique666:


action.yml
```diff
@@ -58,4 +58,7 @@ inputs:
     default: false
 runs:
   using: 'node12'
-  main: 'dist/index.js'
```
Heretique666, line 61: - name: Close Stale Issues
  uses: actions/stale@v3.0.14

PR #263, opened 2021-01-04
Only have issues being closed or marked as stale count as operations

dwelch-r7:
The aim of this PR is to use what I think is a better UX for what "operations per run" means

Currently it's not immediately obvious what counts as an operation unless you go diving into the code, these changes make it so that only either an issue being marked as stale or being closed counts as an operation which simplifies things and makes it easier to rate limit how many issues are affected in a particular run

I'm happy to discuss this though if there are any other ideas out there but this has been a point of frustration for me

### DIFF
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -65,8 +65,8 @@ test('empty issue list results in 1 operation', async () => {
   // process our fake issue list
   const operationsLeft = await processor.processIssues(1);
 
-  // processing an empty issue list should result in 1 operation
-  expect(operationsLeft).toEqual(99);
+  // processing an empty issue list should result in 0 operations
+  expect(operationsLeft).toEqual(100);
 });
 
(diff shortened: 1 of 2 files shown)


hross:
This is tough. The reason we provide this number is because the API is rate limited, not necessarily to limit the number of issues operated on. We want to track how many ops you can do in the API so you know when you might get rate limited or how to restrict the stale bot so it doesn't blow up your API rate limit.

hross (changes requested):
I don't want to change operations left to mean something other than "API operations". Maybe adding an issue limit as a different option would help but I'm not sure how widely desirable that is and I don't want to add more options to the config unless they are used by a lot of people.

hross:
I'm going to close this because I don't think this is the right way to change the number of operations (or at least, think about what that means. Maybe we should clarify the docs.

PR #268, opened 2021-01-16
Remove the stale label when labeled with an exempt one 

### LINKED ISSUE #136: Remove stale label when issue is tagged with exempt label after marking stale
If an exempt label, e.g. `wip` is added **after** the action ran and marked an issue as `stale` it would be nice to remove the `stale` label by the action in the next run. Right now the code seems to only consider exempt labels for skipping but not removing previously labeled issues.

C0ZEN:
**Changes:**

- [x] Prefix private methods with `_`
- [x] Create a new method to remove the stale label (avoid code duplication)
- [x] Before skipping the process when the issue/pr has an exempt label it will check if it has a stale label and remove it if it does (costing one operation)
- [x] Change the syntax to display the default values in italic in the readme so that prettier is no longer messing with them

Closes #136


C0ZEN:
@hross this is the last one from me for this week ;)

C0ZEN:
I think that there is an issue in the tests, I am looking into it.

hross (approved):

Issue #302, opened 2021-02-03
onlyLabels requires all labels to be present


joeveiga:
The `only-labels` option requires all the listed labels to be present in the issue (an AND, not an OR) in order for it to be taken into account. I understand that parameter is sent as is to the api, here: https://github.com/actions/stale/blob/86561461b92875de77a8b2d2e75f004c826e8f45/src/IssueProcessor.ts#L462 
but IMO (and for my use case) it'd be more useful for that option to behave as an OR. Maybe filtering "client side" instead of sending the parameter to the API? Or maybe a whole different config option if we don't want a breaking change?

joeveiga:
Please see #319 . Thanks!

C0ZEN:
@hross this one can be closed since the related PR was merged.

PR #319, opened 2021-02-16
Feat: add any-of-labels option

joeveiga:
#302


src/classes/issues-processor.ts
```diff
@@ -206,6 +206,19 @@ export class IssuesProcessor {
         continue; // don't process exempt issues
       }
 
+      const anyOfLabels: string[] = wordsToList(this.options.anyOfLabels);
+      if (
+        anyOfLabels.length &&
+        !anyOfLabels.some((label: Readonly<string>): boolean =>
+          isLabeled(issue, label)
+        )
+      ) {
```
C0ZEN, line 215: Some is only called if the array is not empty.
So the check regarding the array length seems useless to me.
When the array is empty, it will return `false` which is exactly what you can expect.
```suggestion
      if (!anyOfLabels.some((label: Readonly<string>): boolean =>
          isLabeled(issue, label)
        )
      ) {
```
As a side-note, I think that having a private method would make this easier to read.

action.yml
```diff
@@ -88,6 +88,10 @@ inputs:
     description: 'Only issues or pull requests with all of these labels are checked if stale. Defaults to `[]` (disabled) and can be a comma-separated list of labels.'
     default: ''
     required: false
+  any-of-labels:
+    description: 'Only issues or pull requests with one of these labels are checked if stale. Defaults to `[]` (disabled) and can be a comma-separated list of labels.'
```
C0ZEN, line 92:
```suggestion
    description: 'Only issues or pull requests with one of these labels are checked if stale. Defaults to `` (disabled) and can be a comma-separated list of labels.'
```
It seems that this "lie" about the default value is coming from the current code base and this suggestion can also be made for the existing `only-labels` option.

C0ZEN (commented):
Hello and it's nice to see a contribution!

I have some suggestions (didn't want to add a request changes):
- [x] the README.md is missing some update as well regarding the table containing all the options
- [x] eventually you could also add at the end of the README.md an example regarding this new option to point out the diff with `only-labels`
- [x] it would be better with some coverage to handle at least: 
	- [x] the case were it's empty
	- [x] the case were there is a label but not matching the one in the issue
	- [ ] the case were there is a label matching the one in the issue
	- [ ] the case were there is multiple labels and none match the one in the issue
	- [ ] the case were there is multiple labels and one match the one in the issue

src/classes/issues-processor.ts
joeveiga: @C0ZEN Thanks for the quick review! Well, you're correct, `.some()` does return `false` on an empty array, however in this case we do not want to execute the code inside the `if` block in that case (note the `!`). If we remove the check for empty array then we'd be skipping all issues.

action.yml
joeveiga: > It seems that this "lie" about the default value is coming from the current code base and this suggestion can also be made for the existing `only-labels` option.

Lol yep, that did seem strange... I'll fix it in both of them 😆

src/classes/issues-processor.ts
C0ZEN: Yep sorry, the logic is reversed here because we need to skip if we have some required label and if the current processed issue don't have one of these.
In that case, `includes` make it a little bit more cleaner IMO.
```suggestion
      if (anyOfLabels.length && !anyOfLabels.includes(isLabeled(issue, label))) {
```

joeveiga: @C0ZEN But `.includes()` would be expecting a `string` in this case, and as far as I can tell `isLabeled()` returns a `boolean`.

C0ZEN: Yeah you are right I was thinking of having a fn to get the issue label but it's more work for only a bit of sugar so better stick with this, sorry for the bad suggestions I think it's time for me to hit the bed 😂

README.md
```diff
@@ -29,7 +29,8 @@ Warns and then closes issues and PRs that have had no activity for a specified a
 | `exempt-all-milestones`       | Exempt all issues and PRs with milestones from being marked as stale. (priority over `exempt-milestones` rules) | Optional |
 | `exempt-all-issue-milestones` | Exempt all issues with milestones from being marked as stale. (override `exempt-all-milestones`).               | Optional |
 | `exempt-all-pr-milestones`    | Exempt all PRs with milestones from being marked as stale. (override `exempt-all-milestones`).                  | Optional |
-| `only-labels`                 | Only labels checked for stale issue/PR.                                                                         | Optional |
+| `only-labels`                 | Only issues and PRs with ALL these labels are checked.                                                          | Optional |
+| `any-of-labels`               | Only issues and PRs with ANY of these labels are checked.                                                       | Optional |
```
C0ZEN, line 33:
```suggestion
| `only-labels`                 | Only issues and PRs with ALL these labels are checked. Separate multiple labels with commas (eg. "question,answered").                                                         | Optional |
| `any-of-labels`               | Only issues and PRs with ANY of these labels are checked. Separate multiple labels with commas (eg. "incomplete,waiting-feedback").                                                      | Optional |
```

__tests__/any-of-labels.spec.ts
```diff
@@ -0,0 +1,106 @@
+import {Issue} from '../src/classes/issue';
+import {IssuesProcessor} from '../src/classes/issues-processor';
+import {IIssuesProcessorOptions} from '../src/interfaces/issues-processor-options';
+import {DefaultProcessorOptions} from './constants/default-processor-options';
+import {generateIssue} from './functions/generate-issue';
+
+describe('any-of-labels option', () => {
```
C0ZEN, line 7: I like the fact that you decided to put them in a dedicated file

__tests__/any-of-labels.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  test('should do nothing when not set', async () => {
+    const sut = new IssuesProcessorBuilder()
+      .emptyAnyOfLabels()
+      .issues([{labels: [{name: 'some-label'}]}])
+      .build();
+
+    await sut.processIssues();
+
+    expect(sut.staleIssues).toHaveLength(1);
```
C0ZEN, line 16: I saw that the other tests are not using this method to check the length yet it's the most appropriate one ❤️

__tests__/any-of-labels.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      .issues([{labels: [{name: 'and-this-one'}, {name: 'skip-this-issue'}]}])
+      .build();
+
+    await sut.processIssues();
+
+    expect(sut.staleIssues).toHaveLength(1);
+  });
+});
+
+class IssuesProcessorBuilder {
```
C0ZEN, line 64: I prefer to have all the "tests arrange" in the top of the file.
It makes more sense to me since it's the first thing to code and the first thing executed as well in the tests.
What do you think?

__tests__/any-of-labels.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  private _options: IIssuesProcessorOptions;
+  private _issues: Issue[];
+
+  constructor() {
+    this._options = {...DefaultProcessorOptions};
```
C0ZEN, line 69: Why not assign the properties on creation to reduce the class size instead of using a constructor?

__tests__/any-of-labels.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    this._options.anyOfLabels = labels;
+    return this;
+  }
+
+  emptyAnyOfLabels(): IssuesProcessorBuilder {
+    return this.anyOfLabels('');
+  }
+
+  issues(issues: Partial<Issue>[]): IssuesProcessorBuilder {
+    this._issues = issues.map((issue, idx) =>
```
C0ZEN, line 83:
```suggestion
    this._issues = issues.map((issue, index): Issue =>
```

__tests__/any-of-labels.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      generateIssue(
+        this._options,
+        idx,
+        issue.title || 'Issue title',
+        issue.updated_at || '2000-01-01T00:00:00Z', // we only care about stale/expired issues here
```
C0ZEN, line 88: Using a spread operator would be cleaner IMO.
You can just merge the default values with the given issue.
What do you think?

__tests__/any-of-labels.spec.ts
joeveiga: Yeah I'm still not 100% satisfied with that one cause there is some implementation detail leaking out. It would probably be worth it to explore adding a custom matcher for this in the future, maybe? Like `expect(sut).toHaveProcessedIssues()`.

__tests__/any-of-labels.spec.ts
joeveiga: Well, ideally that fixture would be moved into a separate file and enriched to be shared and reused by all tests. But even if it stays here, I don't really see much value having it on top in this case because 1) the test arrangement would be the instantiation of the builder, not its declaration, 2) the builder itself, albeit useful, is not that interesting, and anyone going through the test suite should be able to get what it does just by looking at how it's used tbh, and 3) having the `describe/test` blocks as close to the beginning of the file as possible is more practical, IMO.

__tests__/any-of-labels.spec.ts
joeveiga: Ah, a OO purist would say the constructor is the place to do it! lol. But really, I'm in the habit of doing it like this for more practical reasons: 1) if you add some more fields, and some of them are initialized via a constructor arg, and others are initialized inline where they are declared, that could get a bit disorganized, 2) it gives a good indication for potential refactorings, e.g. those default options could be injected, instead of imported (imagine we have some other predefined defaults, for example).
But it's just a matter of preference, really. I'm not opposed to getting rid of the constructor 😀.

__tests__/any-of-labels.spec.ts
joeveiga: Sure, that could work. I saw you had that function to generate the mock issues and thought it would be frowned upon if I didn't use it lol. I'm a fan of fluent builder patterns  for this type of setup, maybe adding a `IssueBuilder` in the future can be useful 😉.

C0ZEN (approved):

C0ZEN:
Closes #302

C0ZEN:
@hross if you have the time to check this one and this one #327 🙏🏻 🤟🏻 
Thanks!

C0ZEN:
@joeveiga I gave you some conflicts 😛

hross:
I can merge this when the conflicts are resolved

joeveiga:
@C0ZEN @hross Thanks fellas! Conflicts resolved.

hross (approved):

hross:
I merged a bunch of PRs and will release a new release after all of them have run on master (and possibly one or two more to merge). Probably end of this week and I will cut a release.

PR #336, opened 2021-02-27
feat(only-labels): add 2 new options to distinguish issue and PR configs

### LINKED ISSUE #308: [Feature Request] Separate only-labels into only-issue-labels and only-pr-labels
Example use case that this request is for:

* Only issue marked as "Close?" will be processed.
* But any PR inactive for 150 days will be processed regardless of labels.

There was an unmerged attempt at #206 . This request is a repeat of #219 that was closed by, ironically, the stale bot.

Alternative implementation that would also be acceptable: Have option to disable the action for issue or PR, hence allowing the action to be defined twice with different rules, once for issues and once for PRs.

C0ZEN:
**Changes:**
- [x] Add `only-issue-labels` and `only-pr-labels` options

Closes #308


pllim:
Thank you! 🎉 Which release should I be on the lookout for to have this feature? Is there a timeline for the release?

hross:
expect a new release ~Friday unless you want to test directly off of `main`. we typically test internally for a few days until we can get some real world usage with this action.

PR #337, opened 2021-02-27
feat(statistics): display some stats in the logs

### LINKED ISSUE #262: Add statistics after run
It would be nice to have a basic statistics after the action run, e.g. issues looked at, number of issues marked stale, number of issues no longer stale, number of issues closed etc.

C0ZEN:
**Changes:**
- [x] Add `enable-statistics` option
- [x] Enhance the documentation regarding the debug
- [x] Change the way to mock the `IssuesProcessor` so that it no longers allow to override some methods from the constructor
- [x] Change the way the operations left are calculated to make sure it is made only when a GitHub operation is made (false positives and also missing operations) (only drawback is that the debug cancel before the count)

Closes #262


C0ZEN (commented):
@hross this one is ready and I think it's a good one to close the scope for the next release ;)

Issue #363, opened 2021-03-07
Create a different class between a PR and an issue


C0ZEN:

C0ZEN:
up

C0ZEN:
up

C0ZEN:
up

C0ZEN:
up

C0ZEN:
up

C0ZEN:
up

PR #358, opened 2021-03-05
chore(logs): enhance the logs

C0ZEN:
**Changes:**
- [x] Add logs for the milestones options
- [x] Use `core.setFailed` instead of throwing errors https://github.com/actions/toolkit/tree/main/packages/core#exit-codes
- [x] Tackle the current issues processor logs (partially as a first draft)
- [x] Beautiful logs (partially as a first draft)

Partial of #347 
Closes #361

PR #369, opened 2021-03-08
refactor: extract functionality out of IssueProcessor (WIP)

joeveiga:
## Change comments

- [x] Move some issue specific logic into the `Issue` class.
- [x] Add a `PullRequest` class to clean up all the `if (isPR(..)) { .. } else { .. }` and handle those via polymorphism.
- [x] Add an `IssueFactory` to abstract away the creation of `Issue`/`PullRequest` instances.
- [ ] Create a service to encapsulate the api calls. With that DI, in our tests we can mock the service instead of using the current `IssueProcessorMock` fixture, which is preventing us from testing some logic we currently have in those methods being mocked (see `getLabelCreationDate`, for example).
- [x] Invert dependency on the `Logger`/`IssueLogger` as well (how are you running the test suite with all that spam in the console?! xD).
- [x] Add `NoopLoggerFactory` to be used in our test suite.
- [x] Add a `ChalkLoggerDecorator` decorator class to extract the prettifying logic from the `Logger` class. (the tests are failing because it seems that for jest, a colored string !== the plain string...).

## Motivation

At the moment the `IssueProcessor` class feels a bit bloated.


C0ZEN:
Closing https://github.com/actions/stale/issues/363 then?

joeveiga:
@C0ZEN Oh I didn't see that one... I guess we had the same thought 😅

__tests__/classes/noop-logger-factory.ts
```diff
@@ -0,0 +1,32 @@
+import {Logger} from '../../src/classes/loggers/logger';
+import {LoggerFactory} from '../../src/classes/loggers/logger-factory';
+
+/**
+ * @description Creates Logger instances that won't generate any logs.
+ * Useful for getting rid of verbose output while testing.
+ */
+export class NoopLoggerFactory extends LoggerFactory {
```
C0ZEN, line 8: What do you do in the case where you wish to verify that the logs are ok?

joeveiga: @C0ZEN I guess it depends what you mean with ok. The `Logger` class(es) test suites should catch most logger related bugs. If you want to test something like `it('method f() should print "f() called"', () => ...);` you could inject your own `loggerFactoryMock` with the necessary spies into your SUT. What do you think? What I'm really trying to avoid is having every test case (even the ones not related to logging at all!) spamming the console.

C0ZEN: @joeveiga well I am working on this one https://github.com/actions/stale/pull/358 and as you can see it's mostly logs enhancement.
Having the tests showing the logs as if it's the normal stale action is very useful.
I am not against hiding the logs by default in the tests but IMO we should then have a simple feature like a boolean to switch between between on/off.

C0ZEN: Eventually, you could still provide a mocked class of the Logger calling the real code with `super` but adding an extra feature with a boolean which is `isEnabled` so that we can toggle on/off the logs.
This boolean could be set on the jest env files maybe?

joeveiga: @C0ZEN I see. It looks neat, btw! So, if you have a test suite (or test case) that would benefit from outputting the logs, couldn't you inject the regular `LoggerFactory` for that particular suite/case, instead of the `NoopLoggerFactory`? My only concern with printing to the console by default for the tests is 1) my jest output gets lost in all the noise, and 2) writing to the console is technically (kinda) like writing to a file or a db, that's not a unit test at that point.
IMO a feature to switch logging on/off could be useful if that's a config option we're going to expose to the users, otherwise probably not so much.

C0ZEN: I reply because I dunno if you had the time to read my previous message but:
- Yes, having the logs hell in the test is not good so we can by default get rid of them (all)
- But we need sometimes to enabled them all once again so
- If we replace the Logger class with a NoopLogger in the jest setup https://jestjs.io/docs/en/configuration#globalsetup-string which has a property which can toggle the log it would be very nice IMO

joeveiga: In any case, if you prefer a flag as opposed to the factory we can do that too. Maybe something like:

```ts
export class X {
  constructor(logger: Logger) {}
  // ...
}

// logger
export class Logger {
  constructor(private readonly output: boolean = true) {}
}

// spec file
const x = new X(new Logger(false));
```

src/classes/statistics.ts
```diff
@@ -1,8 +1,9 @@
 import chalk from 'chalk';
 import {Logger} from './loggers/logger';
+import {LoggerFactory} from './loggers/logger-factory';
 
 export class Statistics {
-  private readonly _logger: Logger = new Logger();
+  private readonly _logger: Logger;
```
C0ZEN, line 6: Soon it will be removed but it's ok for now! 😋

src/classes/pull-request.ts
```diff
@@ -0,0 +1,35 @@
+import {Issue} from './issue';
+
+export class PullRequest extends Issue {
```
C0ZEN, line 3: Nice

__tests__/classes/noop-logger-factory.ts
joeveiga: Ah, got it! Yeah that sounds good. Sorry I missed the last message 😄

src/classes/loggers/chalk-logger-decorator.ts
```diff
@@ -0,0 +1,28 @@
+import chalk from 'chalk';
+import {Option} from '../../enums/option';
+import {Logger} from './logger';
+
+/**
+ * @description Pretiffy log messages using chalk api
+ */
+export class ChalkLoggerDecorator extends Logger {
```
C0ZEN, line 8: Since chalk is only related to the "utils" we use right now I would prefer to have a generic name like
```suggestion
export class StylingLoggerDecorator extends Logger {
```
so that if chalk is removed it still have some logic in the naming

src/classes/loggers/issue-logger.ts
```diff
@@ -70,10 +69,12 @@ export class IssueLogger extends Logger {
   }
 
   private _getIssuePrefix(): string {
-    return chalk.red(`[#${this._getIssueNumber()}]`);
+    // return chalk.red(`[#${this._getIssueNumber()}]`);
```
C0ZEN, line 72: Why the comment? WIP?

src/classes/issues-processor.ts
```diff
@@ -698,30 +671,6 @@ export class IssuesProcessor {
     }
   }
 
-  private _getDaysBeforeIssueStale(): number {
```
C0ZEN, line 701: Less code in this class: 😎

src/classes/issue-factory.ts
```diff
@@ -0,0 +1,14 @@
+import {isPullRequest} from '../functions/is-pull-request';
+import {IIssue} from '../interfaces/issue';
+import {IIssuesProcessorOptions} from '../interfaces/issues-processor-options';
+import {Issue} from './issue';
+import {PullRequest} from './pull-request';
+
+export class IssueFactory {
+  constructor(private readonly options: IIssuesProcessorOptions) {}
+  createIssue(issue: Readonly<IIssue>): Issue {
```
C0ZEN, line 9: Ok ok don't hate me for that: an empty row to separate the two? :)

__tests__/classes/noop-logger-factory.ts
C0ZEN: Maybe provide the PR like this and I will take a look later about what can we do to simplify?

C0ZEN (commented):
Small comments but basically it's cool to have a new POV on this.
I am not used to the Factory (pattern?) but for the rest it's nice especially for the split between PR and issue class ;)

src/classes/loggers/issue-logger.ts
joeveiga: Well, my tests were failing cause the colored text was different (?) than the one passed to `.toHaveBeenCalledWith(..)`. Is it just me? That was the main motivation to extract the 'coloring' logic into a decorator class. So that my logger test suite wouldn't break... Since these methods are specific to `IssueLogger`, I didn't want to introduce another `StylingIssueLoggerDecaorator` just for that, so I left them out for now.

C0ZEN:
> Add a ChalkLoggerDecorator decorator class to extract the prettifying logic from the Logger class. (the tests are failing because it seems that for jest, a colored string !== the plain string...).

I fixed it (merged) by using a different chalk configuration for the specs which can be alter if for example we need to test the colours FYI.

PR #383, opened 2021-03-09
feat(remove-stale-when-updated): add 2 options for issues and prs

### LINKED ISSUE #377: Split remove-stale-when-updated for issues and PRs
See https://github.com/actions/stale/issues/373

C0ZEN:
**Changes:**
- [x] Highlight more values in the logs
- [x] Enhance some logs to make them more humanized
- [x] Stop the action sooner right after it was un-stale to avoid trying to close the issue
- [x] Add 2 new options `remove-issue-stale-when-updated` and `remove-pr-stale-when-updated`

Closes #377


C0ZEN (commented):
Merge LGTM

src/classes/issues-processor.ts
```diff
@@ -781,6 +796,26 @@ export class IssuesProcessor {
     return this.options.anyOfLabels;
   }
 
+  private _shouldRemoveStaleWhenUpdated(issue: Issue): boolean {
+    if (issue.isPullRequest) {
```
luketomlinson, line 800: I wonder if we could maybe consolidate this function? There seem to be a lot of `if/else` when we might just be able to return the boolean value the option e.g. `removePrStaleWhenUpdated`. At a minimum, I think it could just be an `else` and no need for the `else if` to check the opposite condition

C0ZEN: I removed two else if, nice catch

luketomlinson (approved):
LGTM @C0ZEN 👍

pllim:
Thank you, all! I appreciate GitHub's continued support for OSS folks.

Issue #373, opened 2021-03-08
Question: Need expert to vet my workflow using this Action & Feature Request: remove-stale-when-updated for issue/PR


pllim:
Hello. I am sorry if I am abusing Issues like this, but I don't see any Discussions set up for this repo and I am unable to understand all nuances of the available features by just looking at README. I am in the process of moving a home grown stale bot to Actions, and I *think* this should do what I want but I am not 100% sure.

```yaml
on:
  schedule:
    # runs every Wednesday at 1am UTC
    - cron: '0 1 * * 3'

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/stale@v3
      with:
        repo-token: ${{ secrets.GITHUB_TOKEN }}
        operations-per-run: 50
        ascending: true

        days-before-issue-stale: -1
        days-before-issue-close: 7
        remove-issue-stale-when-updated: false
        stale-issue-message: 'This issue was labeled as **Close?**. Remove **Close?** label or this will be closed in 7 days.'
        close-issue-message: 'I am going to close this issue because it was marked with **Close?**, but if you feel that this issue should stay open, then feel free to re-open, remove **Close?** and **closed-by-bot** labels, and apply **keep-open** label.'
        close-issue-label: 'closed-by-bot'
        exempt-issue-labels: 'keep-open'
        only-issue-labels: 'Close?'

        days-before-pr-stale: 150
        days-before-pr-close: 30
        remove-pr-stale-when-updated: true
        stale-pr-message: 'This pull request is stale because it has been open 150 days with no activity. Remove **Close?**, comment, or push, or this will be closed in 30 days.'
        close-pr-message: 'I am going to close this pull request as per my previous message, but if you feel that this issue should stay open, then feel free to re-open, remove **Close?** and **closed-by-bot** labels, and apply **keep-open** label.'
        stale-pr-label: 'Close?'
        close-pr-label: 'closed-by-bot'
        exempt-pr-labels: 'keep-open'
```

I am thinking having it run weekly, going through the 50 oldest stuff first (hence, `ascending: true`).

Desired **issue** handling:

* Only comb through open issues marked as "Close?" and then issue `stale-issue-message` and start the clock. Skip issues with "keep-open" label.
* People commenting on the issue will not reset the clock.
* After 7 days (this is separate from the cron setting, though is coincidentally also weekly), issue `close-issue-message`, slap a "closed-by-bot" label, and close the issue.

Desired **pull request** handling:

[...]

PR #390, opened 2021-03-15
feat(remove-stale-when-updated): Split remove-stale-when-updated option into remove-stale-when-updated and remove-stale-when-commented

francoiscampbell:
Currently, although updated PRs are not closed when updated, the stale label is not removed, even after re-running the action:
updated-pr-still-stale
(redacted because this is not my PR)

~This PR also removes the label when the PR is updated (for example by pushing a new commit).~

This PR splits the options to remove the stale by renaming `remove-stale-when-updated` to `remove-stale-when-commented` and introducing a new meaning to `remove-stale-when-updated`, which is to remove the stale label when the stale issue is updated in other ways than comments (typically pushing a new commit).

This also happens to fix (what I think is) a bug that was reported to me today: if an issue has a stale label and then updated without comments, the stale timer isn't reset and the issue is still closed after `days-before-close`, not `days-before-stale + days-before-close` as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ Every argument is optional.
 | `only-pr-labels`              | Only PRs with ALL these labels are checked.<br>Separate multiple labels with commas (eg. "question,answered").<br>_Override `only-labels`_.    |
 | `any-of-labels`               | Only issues and PRs with ANY of these labels are checked.<br>Separate multiple labels with commas (eg. "incomplete,waiting-feedback").         |
 | `operations-per-run`          | Maximum number of operations per run.<br>GitHub API CRUD related.<br>_Defaults to **30**_.                                                     |
-| `remove-stale-when-updated`   | Remove stale label from issue/PR on updates or comments.<br>_Defaults to **true**_.                                                            |
+| `remove-stale-when-commented` | Remove stale label from issue/PR on comments.<br>_Defaults to **true**_.                                                                       |
+| `remove-stale-when-updated`   | Remove stale label from issue/PR on any update, including comments.<br>_Defaults to **true**_.                                                 |
 | `debug-only`                  | Dry-run on action.<br>_Defaults to **false**_.                                                                                                 |
 | `ascending`                   | Order to get issues/PR.<br>`true` is ascending, `false` is descending.<br>_Defaults to **false**_.                                             |
 | `skip-stale-issue-message`    | Skip adding stale message on stale issue.<br>_Defaults to **false**_.                                                                          |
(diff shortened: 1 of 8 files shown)


C0ZEN:
@francoiscampbell I saw this not long ago in the code https://github.com/actions/stale/issues/373#issuecomment-794257153
I do not know if this is an issue or a choice but I think that at the end the action should have a new option to split which kind of trigger should remove the stale (comment or/and update).
And I would prefer to see a release containing this fix with the new option because if the bug is here since a long time people are used to it and basically changing the behaviour could result to unexpected un-stale.
What do you think?

francoiscampbell:
Oh sure. I had originally assumed (we just started using this action yesterday) that the label was an indication of whether the PR was "eligible" for closing or not, and felt that if updating the PR delays the closing, it's not stale anymore and therefore the label should be removed.

It's unfortunate that the current name of the option is `remove-stale-when-updated`, since its behaviour is more like `remove-stale-when-commented`, and `remove-stale-when-updated` is what I would've used for this PR's use-case. Any suggestions for the name?

C0ZEN:
I like the name of the options you propose. It makes sense and you can transform a bug into a fix with a new feature giving more control so I like that.

francoiscampbell:
@C0ZEN Sounds good, I've gone and done the option split. Also updated the the PR title and description.

C0ZEN:
Can you add the option in the readme?

francoiscampbell:
Ah oops, done.

francoiscampbell:
@C0ZEN Do you think this is ready to merge?

C0ZEN (approved):

C0ZEN:
@francoiscampbell conflicts FYI

README.md
```diff
@@ -32,7 +32,8 @@ Every argument is optional.
 | `only-pr-labels`              | Only PRs with ALL these labels are checked.<br>Separate multiple labels with commas (eg. "question,answered").<br>_Override `only-labels`_.    |
 | `any-of-labels`               | Only issues and PRs with ANY of these labels are checked.<br>Separate multiple labels with commas (eg. "incomplete,waiting-feedback").         |
 | `operations-per-run`          | Maximum number of operations per run.<br>GitHub API CRUD related.<br>_Defaults to **30**_.                                                     |
-| `remove-stale-when-updated`   | Remove stale label from issue/PR on updates or comments.<br>_Defaults to **true**_.                                                            |
+| `remove-stale-when-commented` | Remove stale label from issue/PR on comments.<br>_Defaults to **true**_.                                                                       |
+| `remove-stale-when-updated`   | Remove stale label from issue/PR on any update, including comments.<br>_Defaults to **true**_.                                                 |
```
pllim, line 36: Doesn't this make more sense? I can't tell if this is doc typo or intentional as I am not familiar with the code base.

```suggestion
| `remove-stale-when-updated`   | Remove stale label from issue/PR on any update, excluding comments.<br>_Defaults to **true**_.                                                 |
```

T6751:
Can i fix these conflicts and open PR?
@C0ZEN
@francoiscampbell

C0ZEN:
Since everyone is waiting for this I will do it.

PR #429, opened 2021-04-28
chore(release): provide a way to create a local release with less grunt work

### LINKED ISSUE #399: Is there a CHANGELOG file?
I was updating our workflow for the project that I support since we've been using this action for closing out old/abandoned issues and I wanted to see what the updates were from 3.0.14 to 3.0.18 but I don't see a CHANGELOG file.

Do you guys track this somewhere that I'm missing?  If not it would be handy to have that information so I can know if I should upgrade and/or what new features I should look into turning on or if a behaviour changes etc.

Thanks!

C0ZEN:
**Changes:**
- [x] Add `standard-version` as a dependency
- [x] Update the contributing to reflect the local release process

Closes #399


luketomlinson (commented):
So @C0ZEN I've got a few thoughts on this:

1. I feel that highlighting 'releases' in this way makes it seem like most contributors would be tagging and submitting releases, which is usually just a handful of maintainers. 
2. I have mixed feelings about introducing a dependency (standard-version) to do this. 

I wonder if it makes sense to have an npm script that does the `npm run build && nom run pack` similar to what you've added and just explain how that is necessary to build your changes. I just think the tagging/standard-version stuff is not applicable to most people, so maybe just an additional npm script would be sufficient here.  

Thoughts?

C0ZEN:
@luketomlinson not sure to follow you so here is my two cents.

What I would love to see is:
- A changelog (because it's something for me essential, especially nowadays when using bots to auto-upgrade dependencies)
- A release note reflecting the changelog

Obviously, I prefer a changelog that is generated based on commits and as a maintainer, you can clearly have the hands on this since you squash everything. 
You can control beforehand what is the proper commit name to apply.

With that in mind, I proposed at first a PR to use semver and semantic-release through a GitHub action and the existence of a master branch plus an alpha one (optional).
IMO it's neat because it's fully automated, label the issues and PRs released (because one of the issues for consumers is that they saw closed/merged stuff which are not yet in production due to the slow release rate) and can provide alpha release, so it's easier to review the releases on your side (and if consumers want to play).
The type of automation that reduce the actions as a maintainer and benefit everyone.

Anyhow, it was not approved, so I came up with this proposal which only respond to the two major improvements I would love to see.
It's a way for a maintainer to create only a tag based on the commits, update the changelog and the packages.
It is a nice shortcut and if there is somehow an issue with the output you can still alter everything before pushing.
Obviously, it can be achieved manually, but I don't understand why when you can use wonderful tools that reduce the grunt work — so introducing a dependency is required in this sense.

Regarding the contribution file, this is not my wish at all to promote to contributors the run of the release script locally. It should be dedicated for maintainers only, and maybe I should explain it more clearly.

I hope it's more clear to you now and I am looking for a feedback to understand your POV from this. Thanks!

luketomlinson (approved):
@C0ZEN I think we can give this a try. If it ends up being too much complexity we can always simplify.

luketomlinson (approved):

PR #433, opened 2021-05-05
chore(logs): final overhaul of the logs

### LINKED ISSUE #347: Improve the logs
Sometimes, there are some logs that sums up multiple conditions.
It makes them harder to understand when trying to debug without looking into the code.

For example: https://github.com/actions/stale/issues/329#issuecomment-790088583

We could split them to be more meaningful.

Also improving the log for "Fetched issues:" statistics to explain that it's a batch of 100

### LINKED ISSUE #395: improve debug text for `onlyLabels` usage
## Current situation

The grammar for the logging output when you use `onlyLabels` is:

```
[#9184] The option "onlyLabels" was specified to only processed the issues and pull requests with all those labels (2)
```

## Suggestion for improvement

I think we can improve it by saying something like this:

```
[#9184] The option "onlyLabels" was specified so we only process issues and pull requests with all those labels (2)
```

## Related issue(s)

#347

C0ZEN:
**Changes:**
- [x] Display more values in cyan (same way everywhere)
- [x] Display all the options the same way (with links pointing to the readme so that with anchors we could find them easily later)
- [x] Make the logs a little more humanized
- [x] Add more logs especially for the stale conditions

Closes #395 
Closes #347

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Every argument is optional.
 
 | Input                             | Description                                                                                                                                                   |
 | --------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `repo-token`                      | PAT(Personal Access Token) for authorizing repository.<br>_Defaults to **${{ github.token }}**_.                                                              |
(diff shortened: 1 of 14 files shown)


luketomlinson (approved):
🅰️ ➕

PR #440, opened 2021-05-07
Remove unused @octokit/rest and generate dist

luketomlinson:
## Changes
- [x] Remove unused @octokit/rest dependency
- [x] generate dist

## Context

I believe all of the Octokit calls are handled through the @actions/github package


thboop (approved):
LGTM

Issue #414, opened 2021-04-08
Documentation should be updated to not to say all arguments are optional


mattwelke:
I added this action to a repo I maintain (https://github.com/schema-inspector/schema-inspector), using the following:

```yaml
name: 'Close stale issues and PRs'
on:
  schedule:
    - cron: '54 2 * * *'

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v3
```

I chose not to set any arguments because the readme for the action states:

> Every argument is optional.

I figured this meant that there would be a default value for the message to use when it detects a stale issue etc and does its thing, and that I could just add an argument later if I wanted to override this default message.

Instead, when it ran overnight, I saw this in the action logs:

```
Run actions/stale@v3
[#97] Found this pull request last updated 2021-04-07T01:47:53Z
[#97] The option "onlyLabels" was not specified. Continuing the process for this pull request
[#97] Days before pull request stale: 60
[#97] Skipping pull request due to empty stale message
[#71] Found this issue last updated 2021-02-03T23:09:10Z
[#71] The option "onlyLabels" was not specified. Continuing the process for this issue
[#71] Days before issue stale: 60
[#71] Skipping issue due to empty stale message
[#69] Found this issue last updated 2021-04-05T19:36:35Z
[#69] The option "onlyLabels" was not specified. Continuing the process for this issue
[#69] Days before issue stale: 60
[#69] Skipping issue due to empty stale message
[#66] Found this issue last updated 2017-09-25T19:14:33Z
[#66] The option "onlyLabels" was not specified. Continuing the process for this issue
[#66] Days before issue stale: 60
[#66] Skipping issue due to empty stale message
[#62] Found this issue last updated 2017-06-20T18:07:24Z
[#62] The option "onlyLabels" was not specified. Continuing the process for this issue
[#62] Days before issue stale: 60
[#62] Skipping issue due to empty stale message
[#56] Found this issue last updated 2016-11-27T10:08:17Z
[#56] The option "onlyLabels" was not specified. Continuing the process for this issue
[#56] Days before issue stale: 60
[#56] Skipping issue due to empty stale message
[#45] Found this issue last updated 2016-10-28T09:30:31Z
[#45] The option "onlyLabels" was not specified. Continuing the process for this issue
[#45] Days before issue stale: 60
[#45] Skipping issue due to empty stale message
[#40] Found this pull request last updated 2021-04-07T01:11:06Z
[...]

PR #457, opened 2021-05-23
feat(options): simplify config by removing skip stale message options

### LINKED ISSUE #405: `stale-issue/pr-message` required even if `skip-stale-issue/pr-message` is set
`stale-issue-message` / `stale-pr-message` should probably not be required if `skip-stale-issue-message` / `skip-stale-pr-message` is an empty string, since that string never gets used.

Workaround: set  `skip-stale-issue-message` / `skip-stale-pr-message` to a non-empty string.

I think this happens because the check for a commit message (https://github.com/actions/stale/blob/b717aa9f47b3b780bbf93d496adc6f4c465fef82/src/classes/issues-processor.ts#L151-L156) comes before we check if an issue is stale (https://github.com/actions/stale/blob/b717aa9f47b3b780bbf93d496adc6f4c465fef82/src/classes/issues-processor.ts#L265-L269).

Example workflow: https://github.com/neverendingqs/serverless-dotenv-plugin/blob/8c02adbc653a90d681d5ededbb40300677cb8eac/.github/workflows/issues-cron.yml

Example run: https://github.com/neverendingqs/serverless-dotenv-plugin/runs/2209947040?check_suite_focus=true#step:2:35

### LINKED ISSUE #455: Remove `skip-stale-issue-message` and `skip-stale-pr-message` options
## Describe your issue

These options seems useless to me.
We can simply define for example that no message will be sent if `stale-issue-message` is empty.
Also, it's strange that there is no option to cover the close message.

I can provide a PR.

C0ZEN:
## Changes
- [x] Remove the options `skip-stale-issue-message` and `skip-stale-pr-message`
- [x] Change the behaviour to allow to process the issues and PRs when the options `stale-issue-message` and `stale-pr-message` are unset since it should be optional to comment and should not be a reason to exit the workflow
- [x] Change the condition to add a comment when stale by simple checking that the options `stale-issue-message` and `stale-pr-message` are set

## Other

Closes #405
Closes #455

BREAKING CHANGES: remove skip-stale-issue-message and skip-stale-pr-message options. If you used this option, replace it by an empty message for the options stale-issue-message and stale-pr-message


C0ZEN:
@luketomlinson good candidate for a major release!

luketomlinson (approved):
I like it. Make it a little simpler

luketomlinson:
@C0ZEN Some conflicts now ☝️

C0ZEN:
@luketomlinson handled

luketomlinson:
yep @C0ZEN I think we can major version here pretty soon

C0ZEN:
So, what about https://github.com/actions/stale/issues/414?

PR #458, opened 2021-05-23
feat(output): print output parameters

### LINKED ISSUE #452: [Feature] Get the information about the staled issues
When the action has run, it would be nice to get the data of the staled issues with author. This way, you could use the data to use other actions that, for example, send a Slack message to the author.

flaxel:
## Changes
- [x] using the [action output options](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs) to return values

## Context

Closes #452


C0ZEN (approved):

README.md
```diff
@@ -59,6 +59,13 @@ Every argument is optional.
 | `exempt-all-pr-assignees`         | Exempt all PRs with assignees from being marked as stale.<br>_Override `exempt-all-assignees`_.                                                               |
 | `enable-statistics`               | Display some statistics at the end of the logs regarding the stale workflow.<br>Only when the logs are enabled.<br>_Defaults to **true**_.                    |
 
+### List of output options
+
+| Output        | Description                |
+| ------------- | -------------------------- |
+| staled-issues | List of all staled issues. |
```
luketomlinson, line 66: should we call these something more generic to handle the fact that these can be either an `issue` or a `pull request`? I know it's confusing since PR's "are" issues. But maybe just better to call it out so someone isn't surprised to see PRs in the output.

luketomlinson (changes requested):
I love this idea! Left one comment about naming.

README.md
flaxel: So I will add `prs` to the naming so it must be clear.

luketomlinson (approved):
💥 💯

Issue #460, opened 2021-05-24
[Feature] Add label when stale issue gets a response


BenVillalobos:
## The problem
When the issue is marked as stale AND the author responds, the issue will continue to be marked as stale and MSBuild wants to avoid this scenario. See [this example in this repo](https://github.com/actions/stale/issues/288)
## The solution
Add some sort of `add-labels-when-updated-from-stale` and `remove-labels-when-updated-from-stale` to allow this automation:

1. Github action checks only X label for stale-ness
2. After 30 days of inactivity, mark issues with X label as stale
3. Customer responds
4. Action should remove stale label AND remove X label so prevent action from marking as stale again. It could also add some sort of "needs team response" label.

C0ZEN:
@BenVillalobos I agree with this.
The logic of the stale action is mostly to clean up but only when it's logical to do it.
And when the ball is in the maintainers of a given project it's not nice for the author to keep un-stale and un-stale again.

I like the ideas you proposed because it gives a lot of possibility from the consumer side, and it's not altering the main logic of the workflow.
It's just extra optional steps.

Do you wish to provide a PR?

BenVillalobos:
@C0ZEN Sure, could you point me in the right direction of where to add the logic?

C0ZEN:
@BenVillalobos the main logic should be added in https://github.com/actions/stale/blob/main/src/classes/issues-processor.ts.
I think the logic should be after https://github.com/actions/stale/blob/4586dc972d16ceec4afcef82779f2e1cf0293e04/src/classes/issues-processor.ts#L572
As well as
https://github.com/actions/stale/blob/4586dc972d16ceec4afcef82779f2e1cf0293e04/src/classes/issues-processor.ts#L284
Since it's at this moment that the stale label will be removed.

The options should be optionals (if empty string).
Maybe we need to think also about `add-labels-when-updated-from-stale` option which bring new labels and maybe in some cases we should get rid of them. Maybe KISS for now and shall see with time and feature requests.

C0ZEN:
I also think that you need to split the options in half between pr and issue and allow a comma separated list of labels.

Issue #288, opened 2021-01-22
Add a comment and a label on issues and PR released


C0ZEN:
To track easily the merged PR and closed issues without yet a release associated with them, it would be awesome to add a comment and a label on each issue and PR.
The label could only be something like released.
The comment could be more precise to know which release is associated to it.

I can provide a PR and was thinking to use [Semantic Release](https://semantic-release.gitbook.io/semantic-release/) FTW.

See original idea/need here https://github.com/actions/stale/issues/287#issuecomment-765253130

C0ZEN:
😅

C0ZEN:
up

C0ZEN:
Having a bot triggered after a release publish to loop through the related issues and PRs closed to add a label like "released" would help and is a good fit I think.

C0ZEN:
up

C0ZEN:
up

PR #468, opened 2021-05-26
Add support for adding & removing labels when no longer stale

BenVillalobos:
Fixes https://github.com/actions/stale/issues/460

## Changes
Add support for:
- labels-to-remove-when-unstale
- labels-to-add-when-unstale

## Context
In order to prevent scenarios like [this](https://github.com/actions/stale/issues/288) and to reduce the amount of upkeep required for maintainers, allow the action to remove/add labels to prevent marking issues stale over and over again. 

Example of an ideal scenario is this:

1. User submits issue
2. We respond and need author feedback, apply that label
3. Issue goes stale via automation
4. Author (or someone else with the same issue) responds suggesting it's still an issue
5. Automation should remove the `needs-author-feedback` label and add some `author-responded` or `needs-dev-response` label

This allows automation to ONLY run on `needs-author-feedback` issues, or be exempt on `author-responded` issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ Every argument is optional.
 | [remove-stale-when-updated](#remove-stale-when-updated)             | Remove stale label from issues/PRs on updates/comments                   | `true`                |
 | [remove-issue-stale-when-updated](#remove-issue-stale-when-updated) | Remove stale label from issues on updates/comments                       |                       |
 | [remove-pr-stale-when-updated](#remove-pr-stale-when-updated)       | Remove stale label from PRs on updates/comments                          |                       |
-| [labels-to-add-when-updated](#labels-to-add-when-unstale)           | Add specified labels from issue/PR when they become unstale              |                       |
-| [labels-to-remove-when-updated](#labels-to-remove-when-unstale)     | Remove specified labels from issue/PR when they become unstale           |                       |
+| [labels-to-add-when-unstale](#labels-to-add-when-unstale)           | Add specified labels from issue/PR when they become unstale              |                       |
+| [labels-to-remove-when-unstale](#labels-to-remove-when-unstale)     | Remove specified labels from issue/PR when they become unstale           |                       |
 | [debug-only](#debug-only)                                           | Dry-run                                                                  | `false`               |
 | [ascending](#ascending)                                             | Order to get issues/PRs                                                  | `false`               |
 | [start-date](#start-date)                                           | Skip stale action for issues/PRs created before it                       |                       |
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ Every argument is optional.
 | [remove-stale-when-updated](#remove-stale-when-updated)             | Remove stale label from issues/PRs on updates/comments                   | `true`                |
 | [remove-issue-stale-when-updated](#remove-issue-stale-when-updated) | Remove stale label from issues on updates/comments                       |                       |
 | [remove-pr-stale-when-updated](#remove-pr-stale-when-updated)       | Remove stale label from PRs on updates/comments                          |                       |
+| [labels-to-add-when-updated](#labels-to-add-when-unstale)           | Add specified labels from issue/PR when they become unstale              |                       |
+| [labels-to-remove-when-updated](#labels-to-remove-when-unstale)     | Remove specified labels from issue/PR when they become unstale           |                       |
 | [debug-only](#debug-only)                                           | Dry-run                                                                  | `false`               |
 | [ascending](#ascending)                                             | Order to get issues/PRs                                                  | `false`               |
 | [start-date](#start-date)                                           | Skip stale action for issues/PRs created before it                       |                       |
@@ -304,6 +306,20 @@ Default value: unset
 #### remove-pr-stale-when-updated
 
 Override [remove-stale-when-updated](#remove-stale-when-updated) but only to automatically remove the stale label when the pull requests are updated (based on [GitHub issue](https://docs.github.com/en/rest/reference/issues) field `updated_at`) or commented.
+
+Default value: unset
+
+#### labels-to-add-when-unstale
+
+A comma delimited list of labels to add when a stale issue receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
+
+Default value: unset
+
+#### labels-to-remove-when-unstale
+
+A comma delimited list of labels to remove when a stale issue receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
+
+Note: Each label results in a unique API call.
 
 Default value: unset
 
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ Every argument is optional.
 | [remove-stale-when-updated](#remove-stale-when-updated)             | Remove stale label from issues/PRs on updates/comments                   | `true`                |
 | [remove-issue-stale-when-updated](#remove-issue-stale-when-updated) | Remove stale label from issues on updates/comments                       |                       |
 | [remove-pr-stale-when-updated](#remove-pr-stale-when-updated)       | Remove stale label from PRs on updates/comments                          |                       |
-| [labels-to-add-when-unstale](#labels-to-add-when-unstale)           | Add specified labels from issue/PR when they become unstale              |                       |
-| [labels-to-remove-when-unstale](#labels-to-remove-when-unstale)     | Remove specified labels from issue/PR when they become unstale           |                       |
+| [labels-to-add-when-unstale](#labels-to-add-when-unstale)           | Add specified labels from issues/PRs when they become unstale              |                       |
+| [labels-to-remove-when-unstale](#labels-to-remove-when-unstale)     | Remove specified labels from issues/PRs when they become unstale           |                       |
 | [debug-only](#debug-only)                                           | Dry-run                                                                  | `false`               |
 | [ascending](#ascending)                                             | Order to get issues/PRs                                                  | `false`               |
 | [start-date](#start-date)                                           | Skip stale action for issues/PRs created before it                       |                       |
@@ -311,15 +311,15 @@ Default value: unset
 
 #### labels-to-add-when-unstale
 
-A comma delimited list of labels to add when a stale issue receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
+A comma delimited list of labels to add when a stale issue or pull request receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
 
 Default value: unset
 
 #### labels-to-remove-when-unstale
 
-A comma delimited list of labels to remove when a stale issue receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
+A comma delimited list of labels to remove when a stale issue or pull request receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
 
-Note: Each label results in a unique API call.
+Warning: each label results in a unique API call which can drastically consume the limit of [operations-per-run](#operations-per-run).
 
 Default value: unset
 
--- a/__tests__/main.spec.ts
+++ b/__tests__/main.spec.ts
@@ -1255,7 +1255,7 @@ test('stale label should be removed if a comment was added to a stale issue', as
   expect(processor.removedLabelIssues).toHaveLength(1);
 });
 
-test('if specified, labels should be added when unstale', async () => {
+test('when the option "labelsToAddWhenUnstale" is set, the labels should be added when unstale', async () => {
   const opts = {
     ...DefaultProcessorOptions,
     removeStaleWhenUpdated: true,
--- a/__tests__/main.spec.ts
+++ b/__tests__/main.spec.ts
@@ -1256,6 +1256,7 @@ test('stale label should be removed if a comment was added to a stale issue', as
 });
 
 test('when the option "labelsToAddWhenUnstale" is set, the labels should be added when unstale', async () => {
+  expect.assertions(4);
   const opts = {
     ...DefaultProcessorOptions,
     removeStaleWhenUpdated: true,
--- a/__tests__/main.spec.ts
+++ b/__tests__/main.spec.ts
@@ -1255,6 +1255,49 @@ test('stale label should be removed if a comment was added to a stale issue', as
   expect(processor.removedLabelIssues).toHaveLength(1);
 });
 
+test('if specified, labels should be added when unstale', async () => {
+  const opts = {
+    ...DefaultProcessorOptions,
+    removeStaleWhenUpdated: true,
+    addLabelsWhenUnstale: 'test'
+  };
+  const TestIssueList: Issue[] = [
+    generateIssue(
+      opts,
+      1,
+      'An issue that should have labels added to it when unstale',
+      '2020-01-01T17:00:00Z',
+      '2020-01-01T17:00:00Z',
+      false,
+      ['Stale']
+    )
+  ];
+  const processor = new IssuesProcessorMock(
+    opts,
+    async () => 'abot',
+    async p => (p === 1 ? TestIssueList : []),
+    async () => [
+      {
+        user: {
+          login: 'notme',
+          type: 'User'
+        }
+      }
+    ], // return a fake comment to indicate there was an update
+    async () => new Date().toDateString()
+  );
+
+  // process our fake issue list
+  await processor.processIssues(1);
+
+  expect(processor.closedIssues).toHaveLength(0);
+  expect(processor.staleIssues).toHaveLength(0);
+  // Stale should have been removed
+  expect(processor.removedLabelIssues).toHaveLength(1);
+  // Some label should have been added
+  expect(processor.addedLabelIssues).toHaveLength(1);
+});
+
 test('stale label should not be removed if a comment was added by the bot (and the issue should be closed)', async () => {
   const opts = {...DefaultProcessorOptions, removeStaleWhenUpdated: true};
   github.context.actor = 'abot';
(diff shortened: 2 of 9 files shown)


src/classes/issues-processor.ts
```diff
@@ -536,8 +541,10 @@ export class IssuesProcessor {
     issue: Issue,
     staleLabel: string,
     actor: string,
+    addLabelsWhenUpdatedFromStale: Readonly<string>[],
+    removeLabelsWhenUpdatedFromStale: Readonly<string>[],
     closeMessage?: string,
-    closeLabel?: string
+    closeLabel?: string,
```
BenVillalobos, line 547: Nit: remove comma

src/classes/issues-processor.ts
```diff
@@ -911,6 +922,48 @@ export class IssuesProcessor {
...
+      return;
+    }
+
+    const issueLogger: IssueLogger = new IssueLogger(issue);
+
+    issueLogger.info(
+      `Adding labels marked as add-labels-when-updated-from-stale...`
+    );
+
+    this._statistics?.incrementAddedItemsLabel(issue);
```
BenVillalobos, line 958: Not sure if this is the statistic to increment

src/classes/issues-processor.ts
```diff
@@ -571,6 +578,10 @@ export class IssuesProcessor {
     if (this._shouldRemoveStaleWhenUpdated(issue) && issueHasComments) {
       await this._removeStaleLabel(issue, staleLabel);
 
+      // Are there labels to remove or add when an issue is no longer stale?
+      await this._removeLabelsWhenUpdatedFromStale(issue, removeLabelsWhenUpdatedFromStale);
```
BenVillalobos, line 582: Decided to explicitly keep the `FromStale` as part of the name. In theory, some future update could have automation do stuff on _any_ update. This PR is specifically for issues becoming "un-stale"

src/classes/issues-processor.ts
C0ZEN: If you run `npm run all` and no linter is altering this comma, keep it like this it's fine ;)

src/classes/issues-processor.ts
C0ZEN: This is a statistic to reflect a call to GitHub to add (one or multiple - we don't care since it's an array) a label to an issue/PR, so it seems OK to me, but I will have a better insight once I reviewed it.

src/classes/issues-processor.ts
C0ZEN, line 582: It's nice to defer the logic to reduce the complexity of this file.
One thing that could maybe be even better is to completely defer it to a dedicated class/service like milestones and assignees works.

src/classes/issues-processor.ts
```diff
@@ -911,6 +922,48 @@ export class IssuesProcessor {
...
+    labels: Readonly<string>[]
+  ): Promise<void> {
+    if (!labels.length) {
+      return;
+    }
+
+    const issueLogger: IssueLogger = new IssueLogger(issue);
+
+    issueLogger.info(
+      `Removing any labels specified via remove-labels-when-updated-from-stale...`
```
C0ZEN, line 936: There is a better way actually to mention an option (to have a generated link in the logs at the end)
```suggestion
      `Removing all the labels specified via the ${this._logger.createOptionLink(
            Option.RemoveLabelsWhenUpdatedFromStale
          )} option...`
```
You need to update the Option enum as well

src/classes/issues-processor.ts
```diff
@@ -911,6 +922,48 @@ export class IssuesProcessor {
...
+    );
+
+    for (const label of labels.values()) {
+      await this._removeLabel(issue, label);
+    }
+  }
+
+  private async _addLabelsWhenUpdatedFromStale(
```
C0ZEN, line 944: Add tests, please?

src/main.ts
```diff
@@ -76,7 +76,9 @@ function _getAndValidateArgs(): IIssuesProcessorOptions {
     exemptAllAssignees: core.getInput('exempt-all-assignees') === 'true',
     exemptAllIssueAssignees: _toOptionalBoolean('exempt-all-issue-assignees'),
     exemptAllPrAssignees: _toOptionalBoolean('exempt-all-pr-assignees'),
-    enableStatistics: core.getInput('enable-statistics') === 'true'
+    enableStatistics: core.getInput('enable-statistics') === 'true',
+    removeLabelsWhenUpdatedFromStale: core.getInput('remove-labels-when-updated-from-stale'),
```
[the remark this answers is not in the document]
BenVillalobos: Glad you feel the same. My heart says "unstale" even if it isn't proper 😁

src/classes/issues-processor.ts
```diff
@@ -911,6 +922,48 @@ export class IssuesProcessor {
...
+    }
+
+    const issueLogger: IssueLogger = new IssueLogger(issue);
+
+    issueLogger.info(
+      `Removing any labels specified via remove-labels-when-updated-from-stale...`
+    );
+
+    for (const label of labels.values()) {
+      await this._removeLabel(issue, label);
```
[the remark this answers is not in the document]
BenVillalobos: Unless I'm missing something, it doesn't look like there's a single method for that: https://docs.github.com/en/rest/reference/issues#labels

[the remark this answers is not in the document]
C0ZEN: Indeed.
In that case, it would be nice I think to explicitly mention in the readme when describing the option that it will consume one operation per label removal, and it's to use carefully. WDYT?

src/classes/issues-processor.ts
```diff
@@ -911,6 +932,59 @@ export class IssuesProcessor {
...
+    if (!removeLabels.length) {
+      return;
+    }
+
+    const issueLogger: IssueLogger = new IssueLogger(issue);
+
+    issueLogger.info(
+      `Removing all the labels specified via the ${this._logger.createOptionLink(
+        Option.RemoveLabelsWhenUpdatedFromStale
+      )}`
```
C0ZEN, line 948:
```suggestion
      `Removing all the labels specified via the ${this._logger.createOptionLink(
        Option.RemoveLabelsWhenUpdatedFromStale
      )} option`
```

src/classes/issues-processor.ts
```diff
@@ -911,6 +932,59 @@ export class IssuesProcessor {
...
+    );
+
+    for (const label of removeLabels.values()) {
+      await this._removeLabel(issue, label);
+    }
+  }
+
+  private async _addLabelsWhenUpdatedFromStale(
+    issue: Issue,
+    addLabels: Readonly<string>[]
```
C0ZEN, line 958:
```suggestion
    addLabels: ReadonlyArray<string>
```

src/classes/issues-processor.ts
```diff
@@ -911,6 +932,59 @@ export class IssuesProcessor {
...
+    if (!addLabels.length) {
+      return;
+    }
+
+    const issueLogger: IssueLogger = new IssueLogger(issue);
+
+    issueLogger.info(
+      `Removing all the labels specified via the ${this._logger.createOptionLink(
+        Option.AddLabelsWhenUpdatedFromStale
+      )}`
```
C0ZEN, line 969:
```suggestion
      `Removing all the labels specified via the ${this._logger.createOptionLink(
        Option.AddLabelsWhenUpdatedFromStale
      )} option`
```

src/classes/issues-processor.ts
```diff
@@ -911,6 +932,59 @@ export class IssuesProcessor {
...
+    );
+
+    try {
+      this._operations.consumeOperation();
+      this._statistics?.incrementAddedItemsLabel(issue);
+      await this.client.issues.addLabels({
```
C0ZEN, line 975: This specific code should not be executed in dry run mode (`this.options.debugOnly`)

src/classes/issues-processor.ts
```diff
@@ -911,6 +932,59 @@ export class IssuesProcessor {
...
+        owner: context.repo.owner,
+        repo: context.repo.repo,
+        issue_number: issue.number,
+        labels: addLabels
+      });
+    } catch (error) {
+      this._logger.error(
+        `Add labels when updated from stale error: ${error.message}`
```
C0ZEN, line 983:
```suggestion
        `Error when adding labels after updated from stale: ${error.message}`
```

BenVillalobos:
I noticed there was no easy way to check if specific labels were added / removed from issues, so I kept the "make sure some label was added to an issue" test similar to the others, where it just checks that _an_ issue had a label added to it when we unstale it.

It's harder to do that for unstale-issue-has-label-removed without redesigning some of the code, which is more than I'd like to chew. I could add a test that removes a label on unstale and check if the array that contains "issues that had labels removed from them" contains two issues (_removeLabel was called twice and therefore added the same item twice), but it's not ideal.

src/classes/issues-processor.ts
BenVillalobos: See https://github.com/actions/stale/pull/468#issuecomment-854177606, should have replied to this comment.

src/classes/issues-processor.ts
BenVillalobos: Not sure if we can modify the parameter this is passing into, but changing it to a ReadonlyArray gets type complaints: `The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.'

src/interfaces/issues-processor-options.ts
```diff
@@ -47,4 +47,6 @@ export interface IIssuesProcessorOptions {
   exemptAllIssueAssignees: boolean | undefined;
   exemptAllPrAssignees: boolean | undefined;
   enableStatistics: boolean;
+  removeLabelsWhenUnstale: string;
```
luketomlinson, line 50: I don't want to be too picky on naming, but I wonder if a name like `labelsToRemoveWhenUnstale` would read a little better and indicate that this option is a list of labels. To me `removeLabelsWhenUnstale` reads like it should be a `boolean` option. Not a huge deal either way, but curious what your thoughts are on that? @BenVillalobos @C0ZEN

luketomlinson (changes requested):
Good work @BenVillalobos! Let's get the README updated with the new options. I also left a comment about naming.

src/interfaces/issues-processor-options.ts
BenVillalobos: > reads like it should be a boolean option

I agree, and that makes sense considering I followed the naming scheme from `removeIssueStaleWhenUpdated` 😄 will rename.

BenVillalobos:
Just rebased off of main. When running `npm run all` I get some diffs that look to be mostly auto formatting. Should I push them up as a single commit or leave these out?

image

C0ZEN:
@BenVillalobos you need to commit all those changes (probably due to a recent npm i which updated deps due to unpinned dependencies).

README.md
```diff
@@ -43,6 +43,8 @@ Every argument is optional.
 | [remove-stale-when-updated](#remove-stale-when-updated)             | Remove stale label from issues/PRs on updates/comments                   | `true`                |
 | [remove-issue-stale-when-updated](#remove-issue-stale-when-updated) | Remove stale label from issues on updates/comments                       |                       |
 | [remove-pr-stale-when-updated](#remove-pr-stale-when-updated)       | Remove stale label from PRs on updates/comments                          |                       |
+| [labels-to-add-when-unstale](#labels-to-add-when-unstale)           | Add specified labels from issue/PR when they become unstale              |                       |
```
C0ZEN, line 46:
```suggestion
| [labels-to-add-when-unstale](#labels-to-add-when-unstale)           | Add specified labels from issues/PRs when they become unstale              |                       |
```

README.md
```diff
@@ -43,6 +43,8 @@ Every argument is optional.
...
+| [labels-to-remove-when-unstale](#labels-to-remove-when-unstale)     | Remove specified labels from issue/PR when they become unstale           |                       |
```
C0ZEN, line 47:
```suggestion
| [labels-to-remove-when-unstale](#labels-to-remove-when-unstale)     | Remove specified labels from issues/PRs when they become unstale           |                       |
```
For consistency, we use plural

README.md
```diff
@@ -307,6 +309,20 @@ Override [remove-stale-when-updated](#remove-stale-when-updated) but only to aut
 
 Default value: unset
 
+#### labels-to-add-when-unstale
+
+A comma delimited list of labels to add when a stale issue receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
```
C0ZEN, line 314:
```suggestion
A comma delimited list of labels to add when a stale issue or pull request receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
```

README.md
```diff
@@ -307,6 +309,20 @@ Override [remove-stale-when-updated](#remove-stale-when-updated) but only to aut
...
+
+Default value: unset
+
+#### labels-to-remove-when-unstale
+
+A comma delimited list of labels to remove when a stale issue receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
```
C0ZEN, line 320:
```suggestion
A comma delimited list of labels to remove when a stale issue or pull request receives activity and has the [stale-issue-label](#stale-issue-label) or [stale-pr-label](#stale-pr-label) removed from it.
```

README.md
```diff
@@ -307,6 +309,20 @@ Override [remove-stale-when-updated](#remove-stale-when-updated) but only to aut
...
+
+Note: Each label results in a unique API call.
```
C0ZEN, line 322:
```suggestion
Warning: each label results in a unique API call which can drastically consume the limit of [operations-per-run](#operations-per-run).
```

__tests__/main.spec.ts
```diff
@@ -1255,6 +1255,49 @@ test('stale label should be removed if a comment was added to a stale issue', as
   expect(processor.removedLabelIssues).toHaveLength(1);
 });
 
+test('if specified, labels should be added when unstale', async () => {
+  const opts = {
```
C0ZEN, line 1259:
```suggestion
  expect.assertions(4);
  const opts = {
```
IMO it's a good practice for async tests to have this

__tests__/main.spec.ts
```diff
@@ -1255,6 +1255,49 @@ test('stale label should be removed if a comment was added to a stale issue', as
   expect(processor.removedLabelIssues).toHaveLength(1);
 });
 
+test('if specified, labels should be added when unstale', async () => {
```
C0ZEN, line 1258:
```suggestion
test('when the option "labelsToAddWhenUnstale" is set, the labels should be added when unstale', async () => {
```

src/classes/issues-processor.ts
```diff
@@ -944,6 +959,65 @@ export class IssuesProcessor {
     return this.options.removeStaleWhenUpdated;
   }
 
+  private async _removeLabelsWhenUnstale(
+    issue: Issue,
+    removeLabels: Readonly<string>[]
```
C0ZEN, line 964:
```suggestion
    labels: Readonly<string>[]
```

src/classes/issues-processor.ts
```diff
@@ -944,6 +959,65 @@ export class IssuesProcessor {
...
+    );
+
+    for (const label of removeLabels.values()) {
+      await this._removeLabel(issue, label);
+    }
+  }
+
+  private async _addLabelsWhenUnstale(
+    issue: Issue,
+    addLabels: Readonly<string>[]
```
C0ZEN, line 985:
```suggestion
    labels: Readonly<string>[]
```

src/classes/issues-processor.ts
```diff
@@ -944,6 +959,65 @@ export class IssuesProcessor {
...
+  ): Promise<void> {
+    if (!addLabels.length) {
+      return;
+    }
+
+    const issueLogger: IssueLogger = new IssueLogger(issue);
+
+    issueLogger.info(
+      `Removing all the labels specified via the ${this._logger.createOptionLink(
```
C0ZEN, line 994:
```suggestion
      `Adding all the labels specified via the ${this._logger.createOptionLink(
```

src/classes/issues-processor.ts
```diff
@@ -944,6 +959,65 @@ export class IssuesProcessor {
...
+        Option.LabelsToAddWhenUnstale
+      )} option.`
+    );
+
+    this.addedLabelIssues.push(issue);
+
+    if (this.options.debugOnly) {
```
C0ZEN, line 1001: This merge of this one https://github.com/actions/stale/pull/463
This option can now be closer to the GitHub call.
Basically, you can move the if from here to just around "addLabels".
That way, we have a better debugger and we can test the operations per run.

C0ZEN (commented):
Almost done mate!

src/classes/issues-processor.ts
BenVillalobos: Whoops!

C0ZEN (approved):

luketomlinson:
@BenVillalobos some conflicts FYI

dist/index.js
```diff
@@ -299,14 +300,16 @@ class IssuesProcessor {
             else {
                 this._logger.info(`${logger_service_1.LoggerService.yellow('Processing the batch of issues')} ${logger_service_1.LoggerService.cyan(`#${page}`)} ${logger_service_1.LoggerService.yellow('containing')} ${logger_service_1.LoggerService.cyan(issues.length)} ${logger_service_1.LoggerService.yellow(`issue${issues.length > 1 ? 's' : ''}...`)}`);
             }
+            const labelsToAddWhenUnstale = words_to_list_1.wordsToList(this.options.labelsToAddWhenUnstale);
```
BenVillalobos, line 303: Is it reasonable to add this var as a global and have it set once via the constructor? I'm not a fan of changing the args to `processIssue` to pass labels to add/remove when unstale. Not the most familiar with best practices in ts though.

src/classes/issues-processor.ts
BenVillalobos: haven't been ignoring this suggestion. Trying to keep the var's name as `labels` would result in an error passing `labels: labels` when calling the API, presumably because the names were identical.

luketomlinson (approved):
Great work @BenVillalobos!

BenVillalobos:
Thanks @luketomlinson! And thanks for the help shepherding this along @C0ZEN !

jezdez:
Seems like this wasn't enabled in the action.yml: #551

PR #474, opened 2021-06-02
fix(operations): fail fast the current batch to respect the operations limit

### LINKED ISSUE #466: operations-per-run appears to be ignored
I set this to 1:

https://github.com/grpc/grpc-go/runs/2667832758?check_suite_focus=true#step:2:9

Yet 2 issues were marked as stale during this run:

https://github.com/grpc/grpc-go/runs/2667832758?check_suite_focus=true#step:2:117
https://github.com/grpc/grpc-go/runs/2667832758?check_suite_focus=true#step:2:194

C0ZEN:
## Changes

- [x] fix a bad stats related to the consumed operations
- [x] display the stats when the operations per run stopped the workflow
- [x] fail fast the current batch to respect the operations limit, fixes #466
- [x] disable the dry-run for the test by default

### DIFF
--- a/src/classes/issues-processor.ts
+++ b/src/classes/issues-processor.ts
@@ -121,6 +121,11 @@ export class IssuesProcessor {
     }
 
     for (const issue of issues.values()) {
+      // Stop the processing if no more operations remains
+      if (!this.operations.hasRemainingOperations()) {
+        break;
+      }
+
       const issueLogger: IssueLogger = new IssueLogger(issue);
       this._statistics?.incrementProcessedItemsCount(issue);
 
--- a/src/classes/issues-processor.ts
+++ b/src/classes/issues-processor.ts
@@ -417,6 +417,9 @@ export class IssuesProcessor {
           'option which is currently set to'
         )} ${LoggerService.cyan(this.options.operationsPerRun)}`
       );
+      this._statistics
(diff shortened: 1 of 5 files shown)


src/classes/issues-processor.ts
```diff
@@ -121,6 +121,11 @@ export class IssuesProcessor {
     }
 
     for (const issue of issues.values()) {
+      // Stop the processing if no more operations remains
+      if (!this.operations.hasRemainingOperations()) {
+        break;
```
luketomlinson, line 126: maybe a log here so the user knows why issues weren't processed?

C0ZEN: No need, this code will be executed after the break
https://github.com/C0ZEN/stale/blob/2f93a2a49f4c195ae653bb1d88195f1ada50c8f7/src/classes/issues-processor.ts#L407

luketomlinson: ah I see it now. It's just a big loop 😂

luketomlinson (approved):

src/classes/issues-processor.ts
C0ZEN: I am not fond of this because it's not obvious but yeah, one day it would be nice to make the code cleaner!

PR #477, opened 2021-06-03
👷 Update workflows

flaxel:
## Changes

- [x] add [release drafter](https://github.com/release-drafter/release-drafter#--) to create automatically draft release
- [x] add [dependency bot for github actions dependencies](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions) 

## Context

I added Dependabot for GitHub Actions to keep the dependencies up to date as well.
The Release Drafter enables automatic generation of a release, which means it no longer has to be created manually.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,10 @@
 version: 2
 updates:
-  # Enable version updates for npm
   - package-ecosystem: 'npm'
-    # Look for `package.json` and `lock` files in the `root` directory
     directory: '/'
-    # Check the npm registry for updates every day (weekdays)
+    schedule:
+      interval: 'daily'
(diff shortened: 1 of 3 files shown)


C0ZEN:
Pretty sure your PR will not be approved.

I already proposed a PR not along ago to use semantic versioning which was also more complete than this because it allowed to have an alpha release which would be nicer than the current process (issues and PRs are closed, and the docs is updated without release automation, so it's confusing for consumers that refer to the doc reflecting the main branch but are mostly using the release version for their usages...).

luketomlinson:
Going to close this @flaxel. I think we'll stick with the current release process for now.

PR #483, opened 2021-06-06
🎨 Add message grouping

### LINKED ISSUE #475: Utilize toolkits collapsible logs
## The problem
Currently, each issue/PR that is processed has a [#NUMBER] appended to the log line. 
## The solution
I think we can utilize the `core.startGroup` and `core.endGroup` [functionality](https://github.com/actions/toolkit/tree/main/packages/core#logging) in the `IssueLogger` to make these logs a little easier to read.

flaxel:
## Changes

- [x] add message grouping

## Context

Action logs are easier to read.

closes #475


C0ZEN (commented):
Moving to npm 7 is also a turning point and should be reflected in the package to "force" contributors to use this version in particular

package.json
```diff
@@ -44,17 +44,17 @@
     "@types/node": "^15.0.2",
     "@types/semver": "^7.3.5",
     "@typescript-eslint/eslint-plugin": "^4.26.0",
-    "@typescript-eslint/parser": "^4.22.1",
-    "@vercel/ncc": "^0.27.0",
+    "@typescript-eslint/parser": "^4.26.0",
```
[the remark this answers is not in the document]
flaxel: Okay then, I will revert the bumps. I started with the bump because one dependency `@typescript-eslint/parser` was not in sync with the other package `@typescript-eslint/eslint-plugin`.

src/classes/issues-processor.ts
```diff
@@ -121,6 +123,11 @@ export class IssuesProcessor {
     }
 
     for (const issue of issues.values()) {
+      const startMessage = issue.isPullRequest
```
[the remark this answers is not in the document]
flaxel: I will add it.

flaxel:
> Moving to npm 7 is also a turning point and should be reflected in the package to "force" contributors to use this version in particular

Is there a file or something where I have to do this?

luketomlinson:
@flaxel I am going to try to get a major version bump this week since there were some breaking changes in the last couple weeks. 

cc: @C0ZEN, I'm going to try to get to your PRs this week

luketomlinson:
@flaxel Any particular motivation for updating the lock version?

flaxel:
> @flaxel Any particular motivation for updating the lock version?

Not really, I ran the `npm run all` command and I got the update.

src/classes/issues-processor.ts
```diff
@@ -121,6 +124,13 @@ export class IssuesProcessor {
     }
 
     for (const issue of issues.values()) {
+      const startMessage = LoggerService.white(
```
luketomlinson, line 127: I wonder if the start/end group should live in the `IssueLogger`? Then the start group could be part of the initial creation of the logger. 

For the `endGroup` maybe the issueLogger could wrap the `core.endGroup`. 

Also I believe there is a mechanism in the logger for using a `$$type` which will print out either issue or pull request.

luketomlinson (changes requested):
Left a comment. 

I think I'd prefer to not update the `package.lock` in this PR, as it's unrelated to the changes. 

Thanks for implementing this!

C0ZEN:
@flaxel it's because the syntax of the package-lock has a breaking change between npm 6 and 7.
The engines key in the package is used to define which version of node & npm to use FYI.
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines

src/classes/issues-processor.ts
flaxel: However, if I have the `startGroup` in the constructor, then many more groups are created than we actually need, because the `IssueLogger` is always recreated. So I would only create a method `startGroup` and `endGroup` in the logger.

luketomlinson: Isn't an `IssueLogger` created for each issue?

package-lock.json
```diff
@@ -442,15 +442,15 @@
       }
     },
     "@eslint/eslintrc": {
-      "version": "0.4.0",
-      "resolved": "https://registry.npmjs.org/@eslint/eslintrc/-/eslintrc-0.4.0.tgz",
-      "integrity": "sha512-2ZPCc+uNbjV5ERJr+aKSPRwZgKd2z11x0EgLvb1PURmUrn9QNRXFqje0Ldq454PfAVyaJYyrDvvIKSFP4NnBog==",
+      "version": "0.4.2",
```
C0ZEN, line 445: @luketomlinson to avoid this kind of changes maybe it's time to pin the dependencies?

src/classes/issues-processor.ts
```diff
@@ -122,308 +123,313 @@ export class IssuesProcessor {
 
     for (const issue of issues.values()) {
       const issueLogger: IssueLogger = new IssueLogger(issue);
-      this._statistics?.incrementProcessedItemsCount(issue);
+      await issueLogger.grouping(`$$type #${issue.number}`, async () => {
```
luketomlinson, line 126: can you just pass in the `this.processIssue(issue, actor)` instead of wrapping it in another async call? Or does that make Typescript unhappy?

flaxel: Unfortunately, that is not possible as far as I know. We can just push the async call further into the methods.

C0ZEN: It's hardcore to review.
I am not convinced that we really need to have async grouping.
I think it's only useful for parallel processing which is not the case here.
Can we try a basic version and see if the group method without synchronous calls work?

flaxel: I have tried to use the method `startGroup` and `endGroup` so I can remove the `async` for the `grouping` method. But the problem is that the method in the `gouping` method must be `async` because the method `processIssue` is `async`. With this setup I tested it but I got only failed tests and it does not work for me like expected. Maybe you have more success?!

luketomlinson: I think the async is fine in this case. it's clean

luketomlinson (approved):
Nice work @flaxel 🎨 🏅

PR #500, opened 2021-06-08
fix(dry-run): forbid mutations in dry-run

### LINKED ISSUE #499: Dry-run allow to add labels and update the issues :/
## Describe your issue

This PR introduced an issue https://github.com/actions/stale/pull/463 regarding the dry-run.
The project doesn't have test covering the operations per run, so it's only an assumption based on my understanding of the current code.

Here is the lack of dry-run wrapper:

https://github.com/actions/stale/blob/b1da9e1fb185d60a9a18b80f336591de6103121c/src/classes/issues-processor.ts#L743
https://github.com/actions/stale/blob/b1da9e1fb185d60a9a18b80f336591de6103121c/src/classes/issues-processor.ts#L787
https://github.com/actions/stale/blob/b1da9e1fb185d60a9a18b80f336591de6103121c/src/classes/issues-processor.ts#L801

And I think that this block should not be cancelled in a dry-run.
https://github.com/actions/stale/blob/b1da9e1fb185d60a9a18b80f336591de6103121c/src/classes/issues-processor.ts#L821

Sorry!

C0ZEN:
## Changes

- [x] Bring back the dry-run mode by default for the tests to avoid throwing errors
- [x] Forbid to add labels and update the issues in dry-run, fixes #499
- [x] Allow to fetch a pull request in dry-run

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ Every argument is optional.
 | [remove-stale-when-updated](#remove-stale-when-updated)             | Remove stale label from issues/PRs on updates/comments                   | `true`                |
 | [remove-issue-stale-when-updated](#remove-issue-stale-when-updated) | Remove stale label from issues on updates/comments                       |                       |
 | [remove-pr-stale-when-updated](#remove-pr-stale-when-updated)       | Remove stale label from PRs on updates/comments                          |                       |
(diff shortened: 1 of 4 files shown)


luketomlinson (approved):
💥

PR #501, opened 2021-06-09
feat(operations): split the operations-per-run option between queries and mutations

### LINKED ISSUE #495: Feature request: option to limit mutations per run
Forked from #466.

There is an option today that limits the number of operations per run, which limits all github API operations, including reads.  However, I would like an option to limit the number of mutations.

Justification:

1. setting the current option to anything other than "unlimited" (or 9999999) means potentially never making it through _checking_ all the issues/PRs, even if no mutations have been performed.  This means if an issue appears late in the list that does require an action, it may never be updated.

2. allowing the tool to make unlimited changes is problematic in the event of a bug or misconfiguration. For http://github.com/grpc/grpc-go, we very much would like to limit it to something like one ~~operation~~ mutation per hour or so.

cc @C0ZEN

C0ZEN:
## Changes

- [x] Remove the `operations-per-run` option
- [x] Add the `query-operations-per-run` (to read) and `mutation-operations-per-run` (to alter the issues) options, resolves #495
- [x] Split the steps to build/pack/lint/test (easier to debug)

⚠️ this is a breaking change

## Context

For huge projects, it can be scary to process all the issues with this action, especially when it's the first configuration.
The operations-per-run was covering all kind of operations and was helpful, but not enough to process the whole set of issues without changing everything.


C0ZEN:
@luketomlinson up

C0ZEN:
up

src/classes/issues-processor.ts
```diff
@@ -855,7 +845,11 @@ export class IssuesProcessor {
   private async _deleteBranch(issue: Issue): Promise<void> {
     const issueLogger: IssueLogger = new IssueLogger(issue);
 
-    issueLogger.info(`Delete branch from closed $$type - ${issue.title}`);
+    issueLogger.info(`Delete
```
C0ZEN: Undo

C0ZEN:
up

src/classes/issues-processor.ts
```diff
@@ -868,6 +858,27 @@ export class IssuesProcessor {
     }
   }
 
+  private async _getPullRequest(
```
C0ZEN, line 861: To remove (wrong rebase)

C0ZEN:
@luketomlinson would be nice if you could review/merge this one and later create the next major release.
I will work on the refactor to get rid of the “common” options and split the doc between PR and issues once this one is merged to avoid having too many conflicts.

dfawley:
It's sad to see good work go to waste like this.

Issue #470, opened 2021-05-26
Option to ignore new comments on stale issues


jeremystretch:
## The problem
Often when an issue is marked stale, it prompts the OP to post "bump" comments to avoid having the issue marked as stale (undermining Stale's entire purpose). We can set `remove-stale-when-updated` to false to prevent the automatic removal of the stale label, however Stale will not (as far as I can tell) proceed with closing an issue that has received at least one comment since the stale label was applied.

(Our workflow is for a maintainer to review new comments on stale issues, and to manually remove the stale label as appropriate.)

Below is an example:

```
[#5442] Found this issue last updated at: 2021-04-23T19:21:04Z
[#5442] The option only-labels (​https://github.com/actions/stale#only-labels​) was not specified
[#5442] └── Continuing the process for this issue
[#5442] Days before issue stale: 60
[#5442] The issue is not closed nor locked. Trying to remove the close label...
[#5442] There is no close label on this issue. Skip
[#5442] This issue has a stale label
[#5442] The option any-of-labels (​https://github.com/actions/stale#any-of-labels​) was not specified
[#5442] └── Continuing the process for this issue
[#5442] This issue has no milestone
[#5442] └── Skip the milestones checks
[#5442] This issue has no assignee
[#5442] └── Skip the assignees checks
[#5442] This issue is already stale
[#5442] Checking for label on this issue
[#5442] Issue marked stale on: 2021-04-14T04:11:47Z
[#5442] Checking for comments on issue since: 2021-04-14T04:11:47Z
[#5442] Comments not made by actor or another bot: 1
[#5442] Issue has been commented on: true
[#5442] Days before issue close: 30
[#5442] Issue has been updated: false
[#5442] Stale issue is not old enough to close yet (hasComments? true, hasUpdate? false)
[#5442] 1 operation consumed for this issue
```

I believe the relevant logic is [here](https://github.com/actions/stale/blob/e8845990724aed85a03e40a7f08c25673821a934/dist/index.js#L555): Because `hasComments` is true, the issue won't be closed.

## The solution

At a glance, I see two potential solutions:

1. Infer from `remove-stale-when-updated` whether to close issues with both a stale label and comments: If false, assume that we should close a stale issue even if it has a comment.
2. Introduce a new configuration option (e.g. `ignore-stale-comments`) to explicitly control this behavior.

C0ZEN:
I think it's what you need https://github.com/actions/stale/pull/390.
You tell me ;)

[...]

Issue #441, opened 2021-05-08
Stale label not being removed after new activity


Rylan12:
## Describe your issue

The `stale` label isn't being removed from PRs when there's new activity even though `remove-stale-when-updated` isn't set.

## Your stale action configuration

```yml
on:
  push:
    paths:
      - .github/workflows/triage-issues.yml
  schedule:
    # Once every day at midnight UTC
    - cron: "0 0 * * *"
  issue_comment:

jobs:
  stale:
    if: >
      startsWith(github.repository, 'Homebrew/') && (
        github.event_name != 'issue_comment' || (
          contains(github.event.issue.labels.*.name, 'stale') ||
          contains(github.event.pull_request.labels.*.name, 'stale')
        )
      )
    runs-on: ubuntu-latest
    steps:
      - name: Mark/Close Stale Issues and Pull Requests
        uses: actions/stale@v3
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          days-before-stale: 21
          days-before-close: 7
          stale-issue-message: >
            This issue has been automatically marked as stale because it has not had
            recent activity. It will be closed if no further activity occurs.
          stale-pr-message: >
            This pull request has been automatically marked as stale because it has not had
            recent activity. It will be closed if no further activity occurs.
          exempt-issue-labels: "gsoc-outreachy,help wanted,in progress"
          exempt-pr-labels: "gsoc-outreachy,help wanted,in progress"
```

## Further context

I made a comment on https://github.com/Homebrew/brew/pull/11140#issuecomment-835407365 that should have caused the action to remove the `stale` label. The action did run successfully (the logs are available [here](https://github.com/Homebrew/brew/runs/2535097842?check_suite_focus=true)). Here are the relevant lines from the logs:

```
[#11140] Found this pull request last updated 2021-05-08T15:44:35Z
[#11140] The option "onlyLabels" was not specified. Continuing the process for this pull request
[#11140] Days before pull request stale: 21
[#11140] The pull request is not closed nor locked. Trying to remove the close label...
[#11140] There is no close label on this pull request. Skip
[#11140] This pull request has a stale label
[#11140] This pull request has no assignee
[#11140] Skip the assignees checks
[#11140] Found a stale pull request
[#11140] Checking for label on pull request
[#11140] Pull request marked stale on: 2021-05-08T15:44:35Z
[#11140] Checking for comments on pull request since 2021-05-08T15:44:35Z
[#11140] Comments not made by actor or another bot: 0
[...]

Issue #435, opened 2021-05-06
Include option to ignore PR activity


SamirMarin:
### Background
We currently use github actions on one of our gitops repos to autoupdate PRs with the latest main branch. Because of this PRs are updated frequently and considered to have a high amount of activity.

We wanted to also use github actions (actions/stale) to close PRs that had been open for several days. Currently the following is supported by action/stale:

```
with:
          repo-token: ${{ token }}
          stale-pr-message: 'This PR is stale because it has been open for 5 days. Remove stale label or this PR will be closed in 3 days.'
          remove-pr-stale-when-updated: false
          days-before-pr-stale: 5
          days-before-pr-close: 3
```

This works as expected marking any PR as stale (that has not had any activity for 5 days) and closing it 3 days later if still labeled stale. 

This however falls slightly short of our use case. For our use case we would like the option to ignore PR activity and mark the PR as stale if it has been open for 5 days (activity or no activity).

### Request
Add a flag option such as `ignore-pr-activity` that ignores PR activity if set to true and simply tracks the time from when the PR was open.
example:
```
with:
          repo-token: ${{ token }}
          stale-pr-message: 'This PR is stale because it has been open for 5 days. Remove stale label or this PR will be closed in 3 days.'
          remove-pr-stale-when-updated: false
          ignore-pr-activity: true  # defaults to false, when true will count time from the day PR was open ignoring all activity
          days-before-pr-stale: 5
          days-before-pr-close: 3
```

pllim:
I want something similar but I want push event to be counted, but not the comments. So, a stale PR will not be stale anymore when author updates the code, but it will still be stale if something comments on it or move the milestone.

Case study: astropy/astropy#9856 -- The last push was in Jan 2020, well over the stale limit. But we moved the milestone because our workflow is to attach milestones to all PRs. But `actions/stale` thinks it is not stale because milestone was moved in April 2021.

```
2021-05-20T03:18:22.7351365Z [#9856] Found this pull request last updated at: 2021-04-08T22:49:28Z
2021-05-20T03:18:22.7352534Z [#9856] The option only-labels (​https://github.com/actions/stale#only-labels​) was not specified
2021-05-20T03:18:22.7353647Z [#9856] └── Continuing the process for this pull request
2021-05-20T03:18:22.7355561Z [#9856] Days before pull request stale: 150
2021-05-20T03:18:22.7356327Z [#9856] The pull request is not closed nor locked. Trying to remove the close label...
[...]

PR #508, opened 2021-06-15
feat(options): add new options to avoid stale based on comments

C0ZEN:
## Changes

- [x] Add 3 new options to avoid staling when a comment is added
- [x] The scope of `remove-stale-when-updated` will be reduced, resulting in a breaking change (@luketomlinson for the release notes!) when using the option at `false`

⚠️ this is a breaking change

## Context

Helping to close (maybe) #441, #470, #435?
Closes #390 due to no activity


C0ZEN:
I'm kind of stuck with the issue it caused.
May need help...

jeremystretch:
@C0ZEN what's the issue?

C0ZEN:
It makes sense to have this feature but the current implementation is actually helping the case where an issue is marked as stale then still in the same workflow process it for closing process. In such case the freshly staled issue update date is set as today and with the current implementation provided in this PR it will consider the issue as no longer stale if a comment was added. If the new option to install when commented is disabled there is no problem

src/classes/issues-processor.ts
```diff
@@ -615,28 +615,63 @@ export class IssuesProcessor {
...
       issueLogger.info(
-        `Remove the stale label since the $$type has a comment and the workflow should remove the stale label when updated`
+        LoggerService.white('└── '),
+        `Skipping the process since the $$type is now un-stale`
       );
-      await this._removeStaleLabel(issue, staleLabel);
 
       // Are there labels to remove or add when an issue is no longer stale?
       await this._removeLabelsWhenUnstale(issue, labelsToRemoveWhenUnstale);
       await this._addLabelsWhenUnstale(issue, labelsToAddWhenUnstale);
```
flaxel, line 667: There is the error which triggered the failing test. The option `labelsToAddWhenUnstale` is empty and therefore no label will be added.

__tests__/main.spec.ts
```diff
@@ -2352,3 +2425,46 @@ test('processing an issue stale since less than the daysBeforeStale without a st
...
+    async p => (p === 1 ? TestIssueList : []),
+    async (): Promise<IComment[]> => Promise.resolve([]),
+    async () => new Date().toDateString()
+  );
+
+  // process our fake issue list
+  await processor.processIssues(1);
+
+  expect(processor.staleIssues).toHaveLength(1);
+  expect(processor.addedLabelIssues).toHaveLength(1);
```
flaxel, line 2468:
```suggestion

```

C0ZEN:
The last added test is actually failing because just after staling, the workflow continue and then consider the issue as stale but to unstale due to `shouldRemoveStaleWhenUpdated && issueHasUpdate` being both `true`.

C0ZEN:
I close it because I just don't understand the whole workflow.
Each time I change something a test is failing, and actually, I can't define if its bad tests or a bad changes.
The fact that the specs are mocking the label creation date and all the logic to calculate if either or not it should stale or close seems terrible IMO.
What could be nice is to rewrite this whole part — code and specs included — and I just don't have the courage to do so.
Well, free to continue the work from this branch or just tackle this from scratch.

PR #510, opened 2021-06-16
Prep for v4

luketomlinson:
## Changes

- [x] Create dist for changes
- [x] Create changelog

PR #513, opened 2021-06-18
ci(config): add pr title linting

flaxel:
## Changes

- [x] add action to check pull request title using [action-semantic-pull-request](https://github.com/amannn/action-semantic-pull-request)

## Context

The titles of the pull requests are not correct in order to generate the changelog automatically. All types of changes from the [specification](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) are allowed.

https://github.com/actions/stale/pull/510#issuecomment-863097970

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,37 @@
+name: "Linting"
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  title:
+  
+    runs-on: ubuntu-latest
+    
+    steps:
+      - uses: amannn/action-semantic-pull-request@v3.4.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          requireScope: true
+          validateSingleCommit: true
+          scopes: |
+            assignee
+            comment
+            config
+            deps
+            deps-dev
+            dry-run
+            label
+            logs
+            operation
+            option
+            output
+            readme
+            release
+            statistic
+            template
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,4 +1,4 @@
-name: "Linting"
+name: 'PR Check'
 
 on:
   pull_request:
@@ -9,9 +9,7 @@ on:
 
 jobs:
   title:
-  
     runs-on: ubuntu-latest
-    
     steps:
       - uses: amannn/action-semantic-pull-request@v3.4.0
         env:


C0ZEN (approved):
Fine with me regarding one thing: you can use a "!" to declare a breaking change.

.github/workflows/linting.yml
```diff
@@ -0,0 +1,37 @@
...
+  title:
+  
+    runs-on: ubuntu-latest
+    
+    steps:
+      - uses: amannn/action-semantic-pull-request@v3.4.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          requireScope: true
```
C0ZEN, line 20:
```suggestion
          types: |
            fix
            feat
            refactor
            docs
          requireScope: true
```
For this project, we would end up with those as well

C0ZEN: I agree with the requirement of a scope, however, any scope.
And so, based on the documentation, I am not sure that this is possible.

C0ZEN:
Funny because your PR title is not good based on your right ;)

flaxel:
> Funny because your PR title is not good based on your proposition ;)

Why do you think so?

.github/workflows/linting.yml
```diff
@@ -0,0 +1,37 @@
+name: "Linting"
```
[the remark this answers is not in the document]
flaxel: I used the common names to possibly just expand the action. If it needs to be more specific, I would call it `PR Check` and `lint`.

.github/workflows/linting.yml
flaxel: The types from the [specification](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) are automatically allowed. If you want to restrict it, you can define `types`. But then I would also add `chore`.

Yes, that's unfortunately how I understood it, which is why I gave the `types`. But I think that these should be sufficient.

C0ZEN:
The "types" refer to the white-list of types that you can use, right? In that case, you didn't include "ci".

flaxel:
If the types are not declared, all those described in the [specification](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) are taken by default.

.github/workflows/linting.yml
C0ZEN: Ignore my comment then

luketomlinson:
@flaxel @C0ZEN  I am leaning against this and possibly backing out the `standard-version`  changes too. I think it introduces more friction than value, IMO.

C0ZEN:
@luketomlinson since I am not the one making the release, honestly it's whatever you prefer.
The only important point for me is to just have a good changelog because I really think it's helpful and valuable.

luketomlinson:
Thanks for the PR @flaxel. I think I am going to lean way from opinionated stuff like this as I think it introduces more friction than benefit in this case.

flaxel:
No it is okay so there can be a discussion and we can find a solution together. 😁

PR #517, opened 2021-07-01
Make label comparison case insensitive

### LINKED ISSUE #516: Make label comparison case insensitive.
## Describe your issue
The comparison logic for when an issue was labeled stale should not be case sensitive.

luketomlinson:
## Changes

- [x] Fixes #516 

## Context
This can cause some weird behavior for legacy labels. Which makes it seem like it was marked stale much longer ago than it actually was.


flaxel (approved):

src/functions/is-labeled.ts
```diff
@@ -21,6 +21,6 @@ export function isLabeled(
   });
 }
 
-function cleanLabel(label: Readonly<string>): CleanLabel {
+export function cleanLabel(label: Readonly<string>): CleanLabel {
```
C0ZEN, line 24: When I started to reduce the complexity of the main class, I decided to go for one export per file regarding most of the code, including the functions.
Dunno WDYT of this subtle detail, but for consistency I would prefer a dedicated file for this function.

C0ZEN (approved):

PR #522, opened 2021-07-07
docs(stale-issue-comment): update the docs to remove that omitting the option will not send a message

### LINKED ISSUE #521: Option stale-issue-message required?
## Describe your issue

We have omitted the `stale-issue-message` option (in fact, we still have `skip-stale-issue-message` set, but that seems ignored now). According to the docs at https://github.com/actions/stale#stale-issue-message:

> You can skip the comment sending by omitting the option or by passing an empty string.

So it seems since we have omitted the `stale-issue-message` parameter, we should be fine - issues should be marked stale without a comment. However, in our logs, we see  https://github.com/jupyterlab/jupyterlab/runs/2994705101?check_suite_focus=true#step:2:3995 :

>  [#5869] Skipping this issue because it should be marked as stale based on the option days-before-stale (​https://github.com/actions/stale#days-before-stale​) (30) but the option stale-issue-message (​https://github.com/actions/stale#stale-issue-message​) is not set

and the relevant issue is not marked stale. Is giving a `stale-issue-message` mandatory now, with an empty string signifying no message?

## Your stale action configuration

https://github.com/jupyterlab/jupyterlab/blob/cd7ac23f738c9a8792bbb76a256e84a82535446c/.github/workflows/answered.yml

```yml
name: 'Close answered issues'
on:
  schedule:
    - cron: '30 1 * * *'

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v3
        with:
          skip-stale-issue-message: true
          days-before-stale: 30
          days-before-close: 7
          stale-issue-label: 'status:Closing as Answered'
          only-issue-labels: 'status:Answered'

```

C0ZEN:
## Changes

- [x] Update the doc to mention that a message will be sent with the default workflow
- [x] Update the doc to remove the part that say that the stale issue message will not be sent when omitted (since the code is checking for empty string and there is a default value, a comment should be added for issue and PRs due to this default value)
- [x] Add a test to confirm this

Closes #521

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Warns and then closes issues and PRs that have had no activity for a specified a
 
 The configuration must be on the default branch and the default values will:
 
-- Add a label "Stale" on issues and pull requests after 60 days of inactivity
+- Add a label "Stale" on issues and pull requests after 60 days of inactivity and comment them
 - Close the stale issues and pull requests after 7 days of inactivity
 - If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart
 
@@ -175,7 +175,7 @@ Default value: unset
 
 The message that will be added as a comment to the issues when the stale workflow marks it automatically as stale with a label.
 
-You can skip the comment sending by omitting the option or by passing an empty string.
+You can skip the comment sending by passing an empty string.
 
 Default value: unset  
 Required Permission: `issues: write`
@@ -184,7 +184,7 @@ Required Permission: `issues: write`
 
 The message that will be added as a comment to the pull requests when the stale workflow marks it automatically as stale with a label.
 
-You can skip the comment sending by omitting the option or by passing an empty string.
+You can skip the comment sending by passing an empty string.
 
 Default value: unset  
 Required Permission: `pull-requests: write`
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Warns and then closes issues and PRs that have had no activity for a specified a
 
 The configuration must be on the default branch and the default values will:
 
-- Add a label "Stale" on issues and pull requests after 60 days of inactivity and comment them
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -4,7 +4,7 @@ Warns and then closes issues and PRs that have had no activity for a specified a
 
 The default configuration will:
 
-- Add a label "Stale" on issues and pull requests after 60 days of inactivity
+- Add a label "Stale" on issues and pull requests after 60 days of inactivity and comment them
```
luketomlinson, line 7:
```suggestion
- Add a label "Stale" on issues and pull requests after 60 days of inactivity and comment on them
```

src/classes/issues-processor.ts
```diff
@@ -76,6 +76,8 @@ export class IssuesProcessor {
   readonly deletedBranchIssues: Issue[] = [];
   readonly removedLabelIssues: Issue[] = [];
   readonly addedLabelIssues: Issue[] = [];
+  readonly addedStaleCommentIssues: Issue[] = [];
```
luketomlinson, line 79: are these being added just for the test?

luketomlinson: maybe we could just inspect the `statistics._addedIssuesCommentsCount` so we don't have to add a field just for tests?

luketomlinson (changes requested):

src/classes/issues-processor.ts
C0ZEN: Actually, it's consistency regarding all this kind of hacks that are indeed all used to only facilitate the tests.
But yeah, using the stats would be possible.

C0ZEN:
up

luketomlinson (approved):

C0ZEN:
ready

PR #530, opened 2021-07-15
docs(readme): using v4

flaxel:
## Changes

- [x] using v4 for usage examples
- [x] remove unnecessary github token 

## Context

release was published but usage examples were not be edited
default configs just work


luketomlinson:
Ahh good catch @flaxel

luketomlinson (approved):

PR #553, opened 2021-08-25
Create check-dist.yml

brcrista:
This adds a workflow to check the contents of the checked-in `dist/index.js` against the expected version.

See https://github.com/actions/upload-artifact/pull/227 for background and rationale.


luketomlinson (approved):
Thanks @brcrista!

Dragonrebl (commented):
brcrista:brcrista/check-dist

PR #771, opened 2022-07-05
Add option labels-to-remove-when-stale

### LINKED ISSUE #770: Add labels-to-remove-when-stale option
## The problem

We have a workflow where a `deploy` label triggers the deployment of a dedicated test environment on AWS via Terraform (like Heroku's "review apps"). When it's removed, the test environment is destroyed. We'd like to keep this flow, but automatically removed the label when the PR becomes **stale**. Right now, there's only an option to add/remove labels when a PR becomes **unstale**.

## The solution

Please add an option `labels-to-remove-when-stale`.

Basically like the existing [labels-to-remove-when-unstale](https://github.com/actions/stale#labels-to-remove-when-unstale) but triggering when the issue/PR becomes stale (not when it becomes unstale).

Thank you very much for your great work! 🙏

arnodirlam:
## Changes

Adds option `labels-to-remove-when-stale`.

## Context

Resolves #770.


luketomlinson (changes requested):
Thanks for the PR @arnodirlam! 

Just a heads up: You'll need to run `npm run pack` in order to generate the JS file from the TS.

cthomaspdx:
This would be really great option to have.  Seems like the stale bot did too good of a job.

Issue #587, opened 2021-10-11
delete branch is not working due to permission problem


alexInhert:
This is my configuration:

```
# https://github.com/actions/stale
name: 'Close stale issues and PRs'
on:
  schedule:
    - cron: '0 9 * * *'
permissions:
  # All other permissions are set to none
  issues: write
  pull-requests: write
jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v4
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          stale-pr-message: >
            This pull request has been automatically marked as stale because it has not had recent activity.
          days-before-pr-stale: 90
          days-before-pr-close: 5
          exempt-pr-labels: 'pinned'
          remove-pr-stale-when-updated: true
          delete-branch: true
          ascending: true
          operations-per-run: 50
```

The job runs as expected committing on PRs and setting the stale label however the delete-branch doesn't work.
In the log I see:

`�[97m�[34m[#1288]�[39m Error when deleting the branch "�[36fixingissue�[39m" from pull request: Resource not accessible by integration�[39m`

Documentation specify that all is needed is  `pull-requests: write` permission which is set. What is the problem?

alexInhert:
> Maybe simply an issue with the GitHub API [actions/first-interaction#10](https://github.com/actions/first-interaction/issues/10)

I don't think so. I had the problem that listed in that issue for commenting on the PR as well. When I added the `pull-requests: write` it was resolved. I don't think that deleting branch permission is `pull-requests: write` as the documentation in this repo listed. So I think it's an issue either with the documentation here or a bug in the code.

C0ZEN:
@alexInhert without the logs, it's hard to tell.
This option exists since a long time, and I am confident regarding the fact that it's working.
Can you configure your project to show the logs, then paste here an issue case?

The source: https://github.com/actions/stale/blob/cdf15f641adb27a71842045a94023bef6945e3aa/src/classes/issues-processor.ts#L853

alexInhert:
> @alexInhert from what I read, it could be a permission issue (not enough privilege) from the given GitHub token.
> Can you share which permissions you enabled?

As you can see in the example code I shared. I use           `repo-token: ${{ secrets.GITHUB_TOKEN }}`
 which is given by Github. I didn't generate it myself.
 https://docs.github.com/en/actions/security-guides/automatic-token-authentication
 
[...]

PR #796, opened 2022-08-17
docs(readme): add missing permission for delete-branch option in readme

dan-pavlov:
## Changes

- [x] add missing `contents: write` permission to `delete-branch` option in readme

## Context

Updating readme to include a permission that is required for `delete-branch` option that is missing from the readme. Configuring this workflow with that additional permission should prevent issues like https://github.com/actions/stale/issues/587 from being raised again (https://github.com/actions/stale/issues/587 wasn't actually resolved, just closed by a bot due to inactivity).

I have recently implemented this workflow in a repository configured with `delete-branch: true`. After the initial wait period, it started closing stale pull requests, but kept failing when trying to delete corresponding branches with the following error `Error when deleting the branch "BRANCH_NAME" from pull request: Resource not accessible by integration`. 

image

Searching for answers I found the issue https://github.com/actions/stale/issues/587 mentioned above, where someone had the same problem, however no solution was provided there. Looking through the code, I found that [this step](https://github.com/actions/stale/blob/main/src/classes/issues-processor.ts#L920) is executed with the [`deleteRef` action of the Octokit library](https://octokit.github.io/rest.js/v18#git-delete-ref), which in turn makes a request to the following [GitHub REST API delete reference endpoint](https://docs.github.com/en/rest/git/refs#delete-a-reference). According to [GitHub REST API docs, that endpoint requires `contents: write` permission](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-contents).

As the recommended permissions mentioned in [_Stale_ readme](https://github.com/actions/stale#recommended-permissions) are set to `issues: write` and `pull-requests: write`, [any other permissions that are absent from the list are set to `none`](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/#setting-permissions-in-the-workflow), therefore blocking `delete-branch` step from successfully executing. 

After adding `contents: write` permission to the workflow configuration, when _Stale_ workflow closed the next stale pull request, it successfully deleted the corresponding branch.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,12 @@ The configuration must be on the default branch and the default values will:
 ## Recommended permissions
 
 For the execution of this action, it must be able to fetch all issues and pull requests from your repository.  
-In addition, based on the provided configuration, the action could require more permission(s) (e.g.: add label, remove label, comment, close, etc.).  
+In addition, based on the provided configuration, the action could require more permission(s) (e.g.: add label, remove label, comment, close, delete branch, etc.).  
 This can be achieved with the following [configuration in the action](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions) if the permissions are restricted:
 
 ```yaml
 permissions:
+  contents: write # only for delete-branch option
   issues: write
   pull-requests: write
 ```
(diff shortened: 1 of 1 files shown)


davorpa (approved):
Looks reasonable to me

Issue #789, opened 2022-08-07
Make `not_planned` the default close-issue-reason


danielcompton:
## The problem

Currently, everyone who uses actions/stale will need to know to configure the `close-issue-reason` setting to `not_planned` to get the new close type - https://github.blog/changelog/2022-05-19-the-new-github-issues-may-19th-update/

## The solution

Make the default value `'not_planned'` instead of `''`. The reason matches the use in stale - "Won't fix, canceled, duplicate, **stale**"

davorpa:
bump

It have a current PR fullfilling requirements #807

luketomlinson:
Released in `v6`: https://github.com/actions/stale/releases/tag/v6.0.0

PR #807, opened 2022-09-05
fix(options)!: Make `not_planned` the default `close-issue-reason`

david-allison:
## Changes

- [x] ⚠️ **BREAKING CHANGE** `close-issue-reason` defaults to `not_planned` (previously: ``
which defaulted to `completed`) This closes the issue as 'not planned' on GitHub rather than 'completed'
- [x] documentation: minor improvement to 'DefaultProcessorOptions'
   - I'm happy to drop this if it's an issue - unsure on PR requirements for this repo

## Context

GitHub introduced an additional close reason 'Close as not planned':
https://github.blog/changelog/2022-05-19-the-new-github-issues-may-19th-update/

'stale' is a use case for this close reason and makes sense as a default for this action

Fixes https://github.com/actions/stale/issues/789

Feature added: https://github.com/actions/stale/commit/06d2a3904b77e3b34b74319855eebe4627488d19
Updated: https://github.com/actions/stale/commit/aaab997ccea19e4bbf31972dde2a627199681ea4

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Every argument is optional.
 | [close-pr-message](#close-pr-message)                               | Comment on the staled PRs while closed                                      |                       |
 | [stale-issue-label](#stale-issue-label)                             | Label to apply on staled issues                                             | `Stale`               |
 | [close-issue-label](#close-issue-label)                             | Label to apply on closed issues                                             |                       |
-| [close-issue-reason](#close-issue-reason)                           | Reason to use when closing issues                                           |                       |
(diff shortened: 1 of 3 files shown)


david-allison:
Fixed conflicts. Is anything required on my end to get this merged?

jeffwidman (approved):
This seems like the sane default. 

I realize it's technically a breaking change so may require a `v6`, but I doubt many users have workflow that changes base on the `close-issue-reason` attribute so blast radius is probably pretty small... could also argue that the previous value was simply "unset" and this isn't so much a breaking change as simply a "enhancement" of setting the value...

luketomlinson (approved):

Issue #596, opened 2021-10-15
Opposite of exempt-all-assignees


kingthorin:
## The problem
It would be nice if there was a way to only add stale comments on issues/PRs that are actually assigned (to remind people of their commitments).

## The solution
Add an opposite flag like `include-only-assigned: true` (Default `false`).

I apologize if this is somehow already possible and I missed it.

kingthorin:
Any thoughts or input on this?

Have I missed an alternative to accomplish the same?

kingthorin:
Well it would be nice if this was addressed….

kingthorin:
Nope not stale, still functionality that people could use….. :cry:

ushio117:
I don't understand but I want to participate. That's my honest comment.

kingthorin:
No bot still not stale but also still not actioned by project maintainers

C0ZEN:
@kingthorin because there is no maintainer in terms of new features.

kingthorin:
🤷‍♂️Had to add some sort of update for the bot

kingthorin:
Poke, still looking for this.

C0ZEN:
More precisely, would you like the processing to only add stale comments without a stale label?
So that it is never closed?

kingthorin:
Interesting question. I think that's already doable. At least I know my projects have been getting comments/labels without anything being closed. (`days-before-close: -1`) I assume there's a way to skip labeling, I haven't bothered to look.

Mainly I was just looking for a way to reduce spam'ish updates. For most of the projects I work with there's little point in the action/bot updating issues which aren't assigned (like this one). We would rather that it was more of a reminder to assignees that they've got something outstanding which they should be working on.

C0ZEN:
OK, so basically, in other words, you wish to process only issues when there is an assignee?

kingthorin:
Exactly.

C0ZEN:
Willing to do that but on this repo https://github.com/Sonia-corporation/stale/issues/559
I no longer add new features here

kingthorin:
@COZEN thanks for that. We'll have to decide which way to go. I'll discuss with the team. (My hesitation is around fragmentation and upkeep.)

kingthorin:
No stale bot it STILL isn’t stale, we’ll at least not __more__ than this entire repo.

kingthorin:
/le sigh

NO bot still not stale, simply not actioned.

kingthorin:
/le sigh

[...]

PR #817, opened 2022-09-12
Issue 596/include only assigned

JoannaaKL:
## Changes

- Add new boolean option `include-only-assigned`. 

## Context

New option `include-only-assigned` enables users to process only issues/PRs that are already assigned. If there is no assignees and this option is set, issue will not be processed.

Resolves https://github.com/actions/stale/issues/596

### DIFF
--- a/src/classes/issues-processor.ts
+++ b/src/classes/issues-processor.ts
@@ -226,7 +226,7 @@ export class IssuesProcessor {
         `Skipping this $$type because it's assignees list is empty`
       );
       IssuesProcessor._endIssueProcessing(issue);
-      return; // If the issue has an 'include-only-assigned' option, process only issues with nonempty assignees list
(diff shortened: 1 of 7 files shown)


src/classes/issues-processor.ts
```diff
@@ -221,6 +221,14 @@ export class IssuesProcessor {
       return; // Don't process locked issues
     }
 
+    if (this._isIncludeOnlyAssigned(issue)) {
+      issueLogger.info(
+        `Skipping this $$type because it's assignees list is empty`
```
rentziass, line 226:
```suggestion
        `Skipping this $$type because its assignees list is empty`
```

rentziass (approved):
Looks great to me 🚀 I've only pointed what looks to be a typo to me but I could be very wrong 😄

takost (approved):

kingthorin:
Thanks!!

PR #841, opened 2022-10-07
Allow daysBeforeStale options to be float

### LINKED ISSUE #833: Allow daysBeforeStale options to be float
## The problem
Currently we parse the `daysBeforeStale` as integer [here](https://github.com/actions/stale/blob/8e8a0e6680dcd91c625405f57bdac762102e4b7d/src/main.ts#L31). It would be great if we could add non integer values to allow higher resolution comparisons: 0.5days, 1.5days etc. 

## The solution
The solution would be as simple as changing `parseInt` to `parseFloat`. And adjust the messaging in other parts of the code.

irega:
## Changes

- [x] Modifies the initialization of the input options `daysBeforeStale`, `daysBeforeIssueStale` and `daysBeforePrStale` to perform a `parseFloat` instead of `parseInt`.
- [x] Adds new test cases to cover the cases where those params are used with hours instead of full days. 

## Context

Resolves #833

### DIFF
--- a/__tests__/main.spec.ts
+++ b/__tests__/main.spec.ts
@@ -1998,6 +1998,84 @@ test('processing an issue opened since 2 days and with the option "daysBeforeIss
   expect(processor.closedIssues).toHaveLength(0);
 });
 
+test('processing an issue opened since 1 hour and with the option "daysBeforeIssueStale" at 0.1666666667 (4 hours) will not make it stale', async () => {
+  expect.assertions(2);
+  const opts: IIssuesProcessorOptions = {
+    ...DefaultProcessorOptions,
+    daysBeforeStale: 10,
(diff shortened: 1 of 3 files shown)


irega:
@luketomlinson @rentziass @flaxel @JoannaaKL could you take a look at this PR? 🙏

JoannaaKL (approved):

irega:
@JoannaaKL seems the action [failed](https://github.com/actions/stale/actions/runs/3280654571/jobs/5401703561) after merging with the main branch. It's weird as the failed step took 6 hours to be executed, so maybe is a punctual timeout error, could you retry the job? let me know if I can help in some way.

Debu976116 (approved):

Carlexos765 (commented):
Leave a revew

PR #874, opened 2022-12-02
Update how stale handles exempt items

### LINKED ISSUE #766: Stale label gets removed when manually added
## Describe your issue

Currently the `stale` label gets removed again when manually added because of `The issue is no longer stale. Removing the stale label...`

## Your stale action configuration

```yml
jobs:
  stale:
    permissions:
      issues: write  # for actions/stale to close stale issues  
      pull-requests: write  # for actions/stale to close stale PRs
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v5
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          stale-issue-message: 'Marking Issue as stale, will be closed in 7 days if no more activity is seen'
          close-issue-message: 'Closing Issue because it is marked as stale'
          stale-issue-label: 'stale'
          exempt-issue-labels: help wanted,docs,enhancement,feature,parity,discussion,roadmap,bug,breaking,dependencies
          days-before-stale: 30
          days-before-close: 7
          remove-stale-when-updated: true
          enable-statistics: true
```

## Further context

If i have understood the PR #352 correctly, it should have fixed #188 by prioritizing a manually added `stale` label and not remove it again because of the manual add being counted as activity

Current Issue i am trying to set as stale: https://github.com/typegoose/typegoose/issues/499
Current action run with debug enabled: https://github.com/typegoose/typegoose/runs/7060586708
Current Stale action workflow (full file): https://github.com/typegoose/typegoose/blob/b8f4da34e7ccb90108285b97dd99b25b60b48ab0/.github/workflows/stale.yml

Action run Output:
```txt
::group::[#499] Issue #499
[#499] Issue #499
  [#499] Found this issue last updated at: 2022-06-26T11:57:43Z
  [#499] The option only-labels (​[https://github.com/actions/stale#only-labels​)](https://github.com/actions/stale#only-labels%E2%80%8B)) was not specified
  [#499]└── Continuing the process for this issue
  [#499] Days before issue stale: 30
  [#499] The issue is not closed nor locked. Trying to remove the close label...
  [#499]├── The close-issue-label (​[https://github.com/actions/stale#close-issue-label​)](https://github.com/actions/stale#close-issue-label%E2%80%8B)) option was not set
  [#499]└── Skipping the removal of the close label
  [#499] This issue has a stale label
  [#499] An exempt label was added after the stale label.
  [#499] The issue is no longer stale. Removing the stale label...
  [#499] Removing the label "stale" from this issue...
  [#499] The label "stale" was removed
  [#499] Skipping this issue because it has an exempt label
  [#499]1 operation consumed for this issue
  ::endgroup::
```

johnsudol:
## Changes

- ~New option added `remove-stale-from-exempt-items`~
- Updated how stale interacts with exempt items
- Clarified some of the existing documentation

## Context

Originally stale bot would not process exempt items. 

In #136, stale bot was updated to remove stale labels from items with exempt labels. This introduced a problem where if a user wanted to manually add a stale label to an item with an exempt label, the `stale` label would automatically be removed. 

~To resolve this, I've added a flag `remove-stale-from-exempt-items` which removes the `stale` label from an exempt item with recent updates. This comes at the cost of a couple of api calls, so by default this option is set to false.   When the flag is disabled stale will continue to not process exempt items~

~TLDR: Stale bot removed stale labels from exempt items no matter what. The new option `remove-stale-from-exempt-items` allows stale bot to only remove the stale label from exempt issues when they've been recently updated~

This PR reverts #136 and adds additional documentation and logging in order to clarify the behaviour of exempt items

Resolves #766


vanZeben (commented):
Just some language updates that I feel are better, feel free to pick and choose.

Also, we've talked offline about this but just to put the comments in a PR. I personally don't think we should be supporting #136 further and should revert that change entirely. 

There are 2 logical ways to infer what the purpose of this action is;
- The first, in #136, is that the inference is that the `stale` label itself is "owned" by this action. With this in mind, `exempt` means `exempt from usage of the label`. 
- The second is that this action "manages" the `stale` label on issue's/pr's. With this in mind, `exempt` means `exempt from the action managing that issue/pr`. 

Personally I feel that the second inference is more appropriate to the spirit of this action. It also means that we don't have to write edge-cases to actively _prevent_ users from labeling their issue's/pr's, which really leads to a confusing user experience (see #766).

This would mean reverting this PR (which is now adding an additional configuration option to handle edge cases), as well as #136 and being more clear on definitions.

JoannaaKL:
So to paraphrase @vanZeben there are two options:
1. Do not allow the issue/pr to be marked as `stale`, as in [#136](https://github.com/actions/stale/issues/136) 
2. Do not process the issue/pr with the `stale` label at all.
I am also personally in favour of the option number 2 - from my perspective it is more understandable and, as mentioned above, relieves us from supporting all weird edge cases. It'd be also great to explicitly describe it in the docs. (Because if now we're discussing what does _extempt_ mean, imagine someone seeing the plugin for the first time). 😄

__tests__/main.spec.ts
```diff
@@ -1094,42 +1094,89 @@ test('exempt pr labels will not be marked stale', async () => {
...
+      ], // return a fake comment to indicate there was an update
+      async () => new Date().toDateString()
+    );
+
+    // process our fake issue list
+    await processor.processIssues(1);
+
+    expect(processor.staleIssues.length).toStrictEqual(0);
+    expect(processor.closedIssues.length).toStrictEqual(0);
+    expect(processor.removedLabelIssues.length).toStrictEqual(1);
```
pje, line 1137: Thanks for testing both the positive and negative cases! 👏 

I notice the assertions here are slightly different from the ones in the negative case: 

```js
    expect(processor.statistics?.processedIssuesCount).toStrictEqual(1);
    expect(processor.closedIssues).toHaveLength(0);
    expect(processor.removedLabelIssues).toHaveLength(0);
```

**Any reason not to test them the same way here?**

```suggestion
    expect(processor.statistics?.processedIssuesCount).toStrictEqual(1);
    expect(processor.staleIssues).toHaveLength(0);
    expect(processor.closedIssues).toHaveLength(0);
    expect(processor.removedLabelIssues).toHaveLength(1);
```

README.md
```diff
@@ -246,17 +246,17 @@ Required Permission: `pull-requests: write`
 
 #### exempt-issue-labels
 
-The label(s) that can exempt to automatically mark as stale the issues.  
-It can be a comma separated list of labels (e.g: `question,bug`).
+Comma separated list of labels that can be assigned to issues to exclude them from being marked as stale
```
pje, line 249: Nice! Your description is way easier to parse than the previous one.

dist/index.js
```diff
@@ -523,17 +523,22 @@ class IssuesProcessor {
                 }
             }
             if (issue.isStale) {
-                issueLogger.info(`This $$type has a stale label`);
+                issueLogger.info(`This $$type includes a stale label`);
             }
             else {
-                issueLogger.info(`This $$type hasn't a stale label`);
+                issueLogger.info(`This $$type does not include a stale label`);
```
pje, line 529: I appreciate the small grammar improvements. 🥳

src/classes/issues-processor.ts
```diff
@@ -332,16 +332,28 @@ export class IssuesProcessor {
...
+      const isIssueStale =
+        isRemoveStaleFromExemptItemEnabled &&
+        (await this._isIssueStale(issue, staleLabel, staleMessage));
+
+      if (isIssueStale) {
+        issueLogger.info(
+          `The option ${issueLogger.createOptionLink(
+            Option.RemoveStaleFromExemptItem
+          )} is enabled, this $$type is no longer stale`
+        );
```
pje, line 353: Thanks for adding this logging too.

pje (approved):

__tests__/main.spec.ts
johnsudol: Good catch! Totally agree

johnsudol:
I've updated the PR to be aligned with comments above.

Moving forward this will simplify how stale handles exempt items and will allow us to avoid edge cases such as the flag that was previously mentioned in this PR

vanZeben (approved):
LGTM

PR #875, opened 2022-12-05
corrected readme messages

tmarshall:
## Changes

Updated the string messages in the "Configure different stale timeouts but never close a PR" example to match what is defined in the `days-before-X` values.

## Context

The stale PR message was incorrect, in the example. It appears that the string value was copied from the next example, "Configure different stale timeouts." In the example that was corrected, `days-before-stale` is set to `30`, and no PR override is used. But there is a `days-before-pr-close` value of `-1`, which, if I understand the docs correctly, means that PRs will not be closed. Just messaged. 

So I've changed the PR messaging to reflect the `30` days, and removed mention of closing the PR.


tmarshall:
closing due to lack of review

Issue #922, opened 2023-01-25
Action doesn't fail on invalid credentials


neykov:
**Description:**
Upon configuring the action with invalid credentials it reports a successful completion even though it wasn't able to connect.

```
Starting the stale action process...
Error: Get issues for repo error: Bad credentials
No more issues found to process. Exiting...
Statistics:
Operations performed: 1
```

**Action version:**
6.0.1

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
* Configure the action as usual
* Provide invalid `repo-token`

**Expected behavior:**
The job should fail, along with the whole action execution.

**Actual behavior:**
The job successfully completes, even though it's not able to connect.

dsame:
hello @neykov, thanks for your input. We started to investigate the issue

dsame:
The https://github.com/actions/stale/pull/935 merged into the main and will be included it next release.

panticmilos:
hi @neykov, we have released a new major [version](https://github.com/actions/stale/releases/tag/v8.0.0) that is addressing this issue.

Could you confirm it is working as expected?

PR #929, opened 2023-02-02
Track errors during processing

dsame:
**Description:**
Current logic assumes the build is not interrupted on some exception. This leads the build looks succeeded while it is failed in fact.

The PR adds tracking some several exception (currently one) and make the build to fail if some of them have happened. 

**Related issue:**
https://github.com/actions/stale/issues/922

**Check list:**


src/classes/issues-processor.ts
```diff
@@ -556,6 +557,7 @@ export class IssuesProcessor {
         (issue: Readonly<OctokitIssue>): Issue => new Issue(this.options, issue)
       );
     } catch (error) {
+      this.errors.push(error);
       this._logger.error(`Get issues for repo error: ${error.message}`);
       return Promise.resolve([]);
```
IvanZosimov, line 562:
```suggestion
      this.errors.push(error);
      this._logger.error(`Get issues for repo error: ${error.message}`);
      core.setFailed(error.message);
```
In my opinion, if there is a problem with fetching data from the repository, we can simply fail the action. Also it will eliminate the need of any other code that was added to handle errors in this file and in main.ts. Also we can try to add retry logic for fetching, for that reason we need to take a look at Octokit plugin https://github.com/octokit/plugin-retry.js, it will retry fetching several times in case of errors with codes 4xx and 5xx. We can enable it in the line 78 by adding it as the argument to the function `getOctokit(this.options.repoToken);` Check the signature of the `getOctokit()` https://github.com/actions/toolkit/blob/main/packages/github/src/github.ts

PR #935, opened 2023-02-10
abort the build on the error

dsame:
**Description:**
Current logic assumes the build is not interrupted on some exception. This leads the build looks succeeded while it is failed in fact.

The PR makeы the build to fail if some of them have happened.

**Related issue:**
[link to the related issue.](https://github.com/actions/stale/issues/922)

**Check list:**


src/classes/issues-processor.ts
```diff
@@ -74,7 +75,7 @@ export class IssuesProcessor {
 
   constructor(options: IIssuesProcessorOptions) {
     this.options = options;
-    this.client = getOctokit(this.options.repoToken);
+    this.client = getOctokit(this.options.repoToken, {}, retry);
```
IvanZosimov, line 78:
```suggestion
    this.client = getOctokit(this.options.repoToken, undefined, retry);
```
As the second parameter of the function getOctokit() is optional, it means that it has union type: `OctokitOptions | undefined`. Would it be more plain to set it to undefined instead of empty object {}, If we don't specify any Octokit options ?

src/classes/issues-processor.ts
```diff
@@ -556,8 +557,7 @@ export class IssuesProcessor {
         (issue: Readonly<OctokitIssue>): Issue => new Issue(this.options, issue)
       );
     } catch (error) {
-      this._logger.error(`Get issues for repo error: ${error.message}`);
-      return Promise.resolve([]);
+      throw Error(`Get issues for repo error: ${error.message}`);
```
IvanZosimov, line 560:
```suggestion
      throw Error(`Getting issues was blocked by the error: ${error.message}`);
```
Could we also think about changing the rethrown error message to be more meaningful? I'm not sure that the current error message is helpful.

package.json
```diff
@@ -31,14 +31,16 @@
     "stale"
   ],
   "engines": {
-    "node": "12",
-    "npm": "6"
+    "node": "v16",
+    "npm": "v8"
```
IvanZosimov, line 35:
```suggestion
    "node": "16",
    "npm": "8"
```

src/classes/issues-processor.ts
dsame: It is a matter of taste. I prefer to clearly state "we do not pass any options" instead of passing something vague and ambiguous. `undefined` is returning value of some expression, issuing it seems to be weird for me.

src/classes/issues-processor.ts
dsame: I added the orignal backtrace to the re-thrown error

https://github.com/actions/stale/pull/935/files#diff-93e32a46e13f11c2066be36506b0f153d7da2b6b8d787da9177c10692ce50fc5R561

package.json
```diff
@@ -31,14 +31,16 @@
     "stale"
   ],
   "engines": {
-    "node": "12",
-    "npm": "6"
+    "node": "16",
+    "npm": "8"
```
dmitry-shibanov, line 35: I think it's better to put it as `>=7` or `>=7.0.0 <10.0.0-0`

src/classes/issues-processor.ts
dmitry-shibanov: For now with undefined and empty it should be the [same](https://github.com/actions/toolkit/blob/d2b7d85e7cb61467395604aae1b84439c320cb84/packages/github/src/utils.ts#L35). But I would prefer to change it to `undefined` because if default option will be changed it won't get affect for us because an empty object is a truthy value.

src/classes/issues-processor.ts
```diff
@@ -556,8 +557,13 @@ export class IssuesProcessor {
...
       );
     } catch (error) {
-      this._logger.error(`Get issues for repo error: ${error.message}`);
-      return Promise.resolve([]);
+      const rethrow = Error(
+        `Getting issues was blocked by the error: ${error.message}`
+      );
+      rethrow.stack = `${error.stack.split('\n').slice(0, 2).join('\n')}\n${
+        error.stack
+      }`;
```
MaksimZhukov, line 565: Do we really need the stack? In my understanding, the error message is enough to understand what went wrong.

MaksimZhukov (approved):

marko-zivic-93 (approved):

IvanZosimov (approved):

PR #959, opened 2023-03-16
Remove labels on stale

panticmilos:
**Description:**
In the scope of this PR we are introducing possibility of removing labels when issue/PR becomes stale.

**Related issue:**
#770

**Check list:**


action.yml
```diff
@@ -180,6 +180,10 @@ inputs:
     description: 'A comma delimited list of labels to add when a stale issue or pull request receives activity and has the stale-issue-label or stale-pr-label removed from it.'
     default: ''
     required: false
+  labels-to-remove-when-stale:
+    description: 'A comma delimited list of labels to remove when a issue or pull request receives activity and has the stale-issue-label or stale-pr-label added to it.'
```
MaksimZhukov, line 184: I think the following description sounds a little clearer
```suggestion
    description: 'A comma delimited list of labels to remove when an issue or pull request becomes stale'
```

marko-zivic-93 (approved):
I did not notice any new problems except the ones already pointed out

action.yml
IvanZosimov: Agree with @MaksimZhukov, but as I see now, his suggestion was put into the wrong line (should be in the line 184 - description of the `labels-to-remove-when-stale`). If we make such changes in the description of new input, do we need to change description for the similar inputs like: `labels-to-add-when-unstale`, `labels-to-remove-when-unstale` ?

dmitry-shibanov (approved):

MaksimZhukov (approved):

IvanZosimov (approved):

Issue #792, opened 2022-08-11
The action gets stuck checking existing stale PRs without getting to new ones


mezz:
## Describe your issue
I have a long "stale" time of 90 days, and then 30 days to close.
The stale bot uses up all my API actions on iterating over existing stale commits and checking them for comments, and it never gets to apply the stale label to everything that needs it.
I am left with a handful of issues marked stale, but many more need the label.

## Your stale action configuration

https://github.com/mezz/JustEnoughItems/blob/def1bfdcba3af124de6fe78fd23ed39c7f1d41a6/.github/workflows/stale.yml#L14-L28

## Further context

I think the bot needs a strategy that lets it apply stale labels first, and then check random stale commits for their comments. If the action can persist some data between runs, it would be nice for it to track which issue it left off on so that it can start from there on the next run.

steveluscher:
Oh no. And that's with only ~200 issues and ~20 pull requests? I'm definitely going to run into this problem then, having ~1000 issues and ~300 pull requests.

dsame:
Hello @mezz , i thought about different strategies and none of them seems perfect to me. But what do you think about just descending sorting the issues by `updated_at` every time? This way the issues that were not be labeled go first and are updated to be checked next time last. Do you see any disadvantages of this approaches?

mezz:
Hello @dsame ,
It is a bit difficult to predict but I think that may work. 
I can imagine that the oldest stale issues will not get closed with this strategy.
If you can create a sorting option, I am happy to try it out and let you know if things improve though.

Are you able to store any state between runs, so that it can "resume" when it runs out of actions?

jklymak:
We are discussing trying this in Matplotlib (>1600 open issues and > 300 open PRs).  It's a little unclear what "operations-per-run" refers to - will we be making >1900 operations just to get the state of each Issue/PR?  Or is it 20 operations in batches of 100?  After that, is each label/close an operation?  Sorry if this should be clear - not super experienced with what GH Actions consider a call or not....  Note that https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions claims Actions get 1000/h, so maybe we don't need to limit to anything as low as 30...  Thanks for any help; hopefully these clarifications are not hijacking this issue.

dsame:
@mezz in theory it is possible to use the same storage that is used for caches to store the state but it will require extra API calls just to "get into the previous state" - that i want to avoid.

[...]

PR #1033, opened 2023-06-15
Optimize API calls

dsame:
**Description:**

Each next run should not proceed the issues proceed during the previous run. In the other case running from the very beginning never reach the issues beyond API limit.

Changes:
 1. removed unnecessary extra call to getPullRequest API, this implied adding isDraft property and refactoring unit tests
 1. Introduced and implemented IState interface with `persist`, `rehydrate` and `reset` method
   - rehydrate is called before the processing
   - persist is called after the processing
   - reset when all the issues are processed and the next run should be started from the beginning
 1. Every issue is marked as processed with IState.isIssueProcessed method and it is skipped during the processing if IState.isIssueProcessed returns true for it
 
 Implementation details:
  - getStateInstance introduced for mocking during testing
  - persistence implemented with @actions/http-client (to store state) and octokit (to remove state)

**Related issue:**
[link to the related issue.](https://github.com/actions/stale/issues/792)

**Check list:**


src/classes/state/state-cache-storage.ts
```diff
@@ -0,0 +1,52 @@
...
+      core.warning(
+        `Saving the state was not successful due to "${
+          error.message || 'unknown reason'
+        }"`
+      );
+    }
+  }
+
+  async restore(): Promise<string> {
+    const tmpDir = fs.mkdtempSync('state-');
```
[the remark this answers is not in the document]
IvanZosimov: Update: as `state-` is used as a prefix for the dir name this question is resolved.

src/classes/state/state.ts
```diff
@@ -0,0 +1,71 @@
+import {IState} from '../../interfaces/state/state';
+import * as core from '@actions/core';
+import {IIssuesProcessorOptions} from '../../interfaces/issues-processor-options';
+import {IStateStorage} from '../../interfaces/state/state-storage';
+import {IIssue} from '../../interfaces/issue';
+
+export type IssueID = number;
```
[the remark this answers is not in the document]
dsame: I want to hide the exact type of because of

1. at the moment i am  not 100% sure `issue.number` is the robust unambiguous issue ID, so there's chance the other issue property will be selected as ID and it can happen to be of some other type. Aliasing the type makes the change painless. This is the 1st benifit. 

2. named type improves readability Set<IssueID> - clearly says about "a set of issue identificator" while Set<number> is nothing but just "a set of numbers of unclear purpose". It works especially good with functions parameters.

src/classes/actions-cache/download.ts
```diff
@@ -0,0 +1,65 @@
+import {createHttpClient, getCacheApiUrl} from './http-client';
+import {retryTypedResponse} from './retry';
+import {isSuccessStatusCode} from './http-responses';
+import {HttpClient} from '@actions/http-client';
+import {downloadCacheHttpClient} from '@actions/cache/lib/internal/downloadUtils';
```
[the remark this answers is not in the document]
IvanZosimov: Also using restoreCache() and cacheSave() functions will help to simplify code greatly.

src/classes/state/state-cache-storage.ts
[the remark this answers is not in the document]
dsame: according to the [docs](https://nodejs.org/docs/latest-v16.x/api/fs.html#fsmkdtempprefix-options-callback) mkdtempSync creates an unique temporary directory

src/classes/actions-cache/download.ts
```diff
@@ -0,0 +1,65 @@
...
+) => {
+  const httpClient = createHttpClient();
+  const archiveUrl = await getCacheArchiveUrl(
+    httpClient,
+    cacheKey,
+    cacheVersion
+  );
+
+  if (!archiveUrl) {
+    return undefined;
```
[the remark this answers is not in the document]
dsame: The warning is issued here: https://github.com/actions/stale/pull/1033/files#diff-25b189a80bad839ebb6267eeb30275c1238170b6aa8d802f5d1fa720ae2f82a2R36 because it is not specific to the used storage but it is a common for any method to persist state.

Meanwhile the imported `cache` code provides enough debug info to describe the exact problem with actions cache

src/classes/actions-cache/download.ts
```diff
@@ -0,0 +1,65 @@
...
+  core.debug(`getCacheEntry response is:\n${JSON.stringify(cacheResult)}`);
+  const cacheDownloadUrl = cacheResult?.archiveLocation;
+  if (!cacheDownloadUrl) {
+    // Cache archiveLocation not found. This should never happen, and hence bail out.
+    throw new Error('Cache not found.');
+  }
+  return cacheDownloadUrl;
+};
+
+export const downloadFileFromActionCache = async (
```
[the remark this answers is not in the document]
dsame: There are 2 reasons to keep it:

 It is a brand name of the service used to persist the state, it is not "a some cache" and using a word cache makes a question "how a caching (usually temporary) does relate to storing (usually permanent)" 
 
 I am not 100% sure at the moment the actions cache is the perfect solution. It is possible some day it will be decided to replace it with come cloud provider (say 90 days lifetime will be not enough)

src/classes/issues-processor.ts
```diff
@@ -561,7 +573,8 @@ export class IssuesProcessor {
       this.statistics?.incrementFetchedItemsCount(issueResult.data.length);
 
       return issueResult.data.map(
-        (issue: Readonly<OctokitIssue>): Issue => new Issue(this.options, issue)
+        (issue): Issue =>
+          new Issue(this.options, issue as Readonly<OctokitIssue>)
```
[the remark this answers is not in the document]
dsame: The previous calls `this.client.rest.issues.listForRepo` and `OctokitIssue` uses the slightly different definitions of the underlining rest response schemas:

 The first: `state_reason?: string | null;`
 The second: `state_reason?: "completed" | "reopened" | "not_planned" | null;`
 
 Having `(issue: Readonly<OctokitIssue>)` produces compilation error. The exact reason of the problem is unknown.

src/classes/state/state-cache-storage.ts
```diff
@@ -0,0 +1,52 @@
...
+    const fileName = path.join(tmpDir, STATE_FILE);
+    try {
+      await downloadFileFromActionCache(fileName, CACHE_KEY, CACHE_VERSION);
+      if (!fs.existsSync(fileName)) {
+        core.info(
+          'There is no state persisted, probably because of the very first run or previous run failed'
```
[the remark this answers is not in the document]
dsame: done

src/classes/state/state.ts
```diff
@@ -0,0 +1,71 @@
...
+      core.warning('The state is not persisted in the debug mode');
+      return;
+    }
+    core.info(
+      `state: persisting info about ${this.processedIssuesIDs.size} issue(s)`
+    );
+    return this.stateStorage.save(this.serialized);
+  }
+
+  async rehydrate(): Promise<void> {
```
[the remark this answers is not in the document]
dsame: done

src/classes/actions-cache/retry.ts
```diff
@@ -0,0 +1,127 @@
...
+
+const sleep = (milliseconds: number): Promise<void> =>
+  new Promise(resolve => setTimeout(resolve, milliseconds));
+// The default number of retry attempts.
+const DefaultRetryAttempts = 2;
+// The default delay in milliseconds between retry attempts.
+const DefaultRetryDelay = 5000;
+
+const retry = async <T>(
+  name: string,
```
nikolai-laevskii, line 32: Let's represent parameters with an object when there're so many of them. There's no way one would remember this order. Same goes for other functions in this file.

src/classes/actions-cache/upload.ts
```diff
@@ -0,0 +1,180 @@
...
+import {TypedResponse} from '@actions/http-client/lib/interfaces';
+import {ReserveCacheError, ValidationError} from '@actions/cache';
+import {isSuccessStatusCode} from './http-responses';
+import {retryHttpClientResponse, retryTypedResponse} from './retry';
+import {getOctokit} from '@actions/github';
+import {retry as octokitRetry} from '@octokit/plugin-retry';
+import {createHttpClient, getCacheApiUrl} from './http-client';
+
+const uploadFile = async (
+  httpClient: HttpClient,
```
nikolai-laevskii, line 13: Again, too many parameters, let's stick to 3 max and when there's more, start representing them with object.

src/classes/actions-cache/upload.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  const additionalHeaders = {
+    'Content-Type': 'application/octet-stream',
+    'Content-Range': contentRange
+  };
+
+  const resourceUrl = getCacheApiUrl(`caches/${cacheId.toString()}`);
+  const fd = fs.openSync(filePath, 'r');
+  try {
+    const uploadChunkResponse = await retryHttpClientResponse(
```
nikolai-laevskii, line 36: Let's move it to the separate function, it's a bit hard to read with all the wraps around. Seems like `uploadChunk` would be an appropriate name for such function.

src/classes/actions-cache/upload.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    if (!isSuccessStatusCode(uploadChunkResponse.message.statusCode)) {
+      throw new Error(
+        `Cache service responded with ${uploadChunkResponse.message.statusCode} during upload chunk.`
+      );
+    }
+  } finally {
+    fs.closeSync(fd);
+  }
+  return;
```
nikolai-laevskii, line 68: Redundant return statement.

src/classes/actions-cache/upload.ts
```diff
@@ -0,0 +1,180 @@
...
+  };
+  const response = await retryTypedResponse('reserveCache', async () =>
+    httpClient.postJson<{cacheId: number}>(
+      getCacheApiUrl('caches'),
+      reserveCacheRequest
+    )
+  );
+
+  // handle 400 in the special way
+  if (response?.statusCode === 400)
```
nikolai-laevskii, line 110: Shouldn't we check response as well for that specific handler? 400 is a general error and it may not always correspond to the case described here.

src/classes/actions-cache/upload.ts
```diff
@@ -0,0 +1,180 @@
...
+    );
+
+    await uploadFile(httpClient, cacheId, filePath, fileSize);
+
+    await commitCache(httpClient, cacheId, fileSize);
+  } catch (error) {
+    const typedError = error as Error;
+    if (typedError.name === ValidationError.name) {
+      throw error;
+    } else if (typedError.name === ReserveCacheError.name) {
```
nikolai-laevskii, line 174: `else` is pointless here as early return (in form of throw) is implemented in first clause. Perhaps we can take advantage of early return further down the chain as well.

src/classes/actions-cache/upload.ts
```diff
@@ -0,0 +1,180 @@
...
+  // TODO: better way to get repository?
+  const repo = process.env['GITHUB_REPOSITORY'];
+  core.debug(`remove cache "${cacheKey}"`);
+  try {
+    // TODO: replace with client.rest.
+    await client.request(
+      `DELETE /repos/${repo}/actions/caches?key=${cacheKey}`
+    );
+  } catch (error) {
+    if (error.status) {
```
nikolai-laevskii, line 83: Shouldn't we check for specific status?

nikolai-laevskii (changes requested):
It seems to be working as it should so far. And the rest of the code seems ok stylewise.

src/classes/actions-cache/http-client.ts
```diff
@@ -0,0 +1,44 @@
+import {HttpClient} from '@actions/http-client';
```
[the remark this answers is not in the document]
dsame: done

src/classes/state/state.ts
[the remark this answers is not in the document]
dsame: You are right, REST response have `id: integer64` and `node_id:string` properties, but we use `IIssue` interface wrapping the response that does not have neither `id` nor `node_id` fields at the moment. Adding the to interface cause great impact on the tests codebase and it is not necessary for the action intended to work within the same repo.

But, if some problem are reported the fix is trivial: modify `IssueID` alias, `State.addIssueToProcessed` and `State.isIssueProcessed` method implementation.

src/classes/actions-cache/http-responses.ts
```diff
@@ -0,0 +1,19 @@
+import {TypedResponse} from '@actions/http-client/lib/interfaces';
+import {HttpClientError} from '@actions/http-client';
+
+export const isSuccessStatusCode = (statusCode?: number): boolean => {
```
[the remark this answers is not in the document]
dsame: The code is inherited from the `@toolkit/cache` and i prefer to keep it unchanged because the API is undocumented and i'd prefer to avoid the risks. The whole folder `actions-cache` is the subject of the refactoring anyway

src/classes/actions-cache/retry.ts
dsame: It is inherited code as well. It has its own code-styling and motivations. I prefer to keep it as is until the refactoring.

src/classes/actions-cache/upload.ts
dsame: done

src/classes/actions-cache/upload.ts
dsame: It is inherited code and investigating the undocumented API is out of scope this issue

src/classes/actions-cache/upload.ts
dsame: done

src/classes/actions-cache/upload.ts
dsame: There's no too much info about the possible status and since `octokit` is high-level wrapper i expect the thrown error object contains all the necessary info

src/classes/actions-cache/upload.ts
dsame: Well, then `uploadFile` will have nothing but call of `uploadChunk` and become useless. 

I improved readability a bit by placing opening a stream in its own function instead.

src/classes/actions-cache/upload.ts
dsame: well, the params does not seem to be related and their are unified through the module. There's no restrictions through the whole project so i prefer to leave it as is untill the whole repo refactoring with [eslint](https://eslint.org/docs/latest/rules/max-params) rule turned on.

src/classes/state/state.ts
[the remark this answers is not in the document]
nikolai-laevskii: Let's refer to it as number then, not ID.

src/classes/actions-cache/upload.ts
nikolai-laevskii: Alright, this way is fine too.

nikolai-laevskii (approved):

dusan-trickovic (approved):
Everything looks alright to me :) Tested a workflow run using this branch and everything works error-free. The comments I left below are just nitpicks and can be ignored, as there are no problems present. Thank you!

src/classes/state/state-cache-storage.ts
```diff
@@ -0,0 +1,99 @@
...
+  fs.mkdirSync(tmpDir, {recursive: true});
+  return tmpDir;
+};
+
+const unlinkSafely = (filePath: string) => {
+  try {
+    fs.unlinkSync(filePath);
+  } catch (foo) {
+    /* ignore */
+  }
```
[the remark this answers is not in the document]
dsame: It must, i ignore the exception thrown if the file does not exist

src/classes/state/state-cache-storage.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+const resetCacheWithOctokit = async (cacheKey: string): Promise<void> => {
+  const token = core.getInput('repo-token');
+  const client = getOctokit(token, undefined, octokitRetry);
+  // TODO: better way to get repository?
+  const repo = process.env['GITHUB_REPOSITORY'];
+  core.debug(`remove cache "${cacheKey}"`);
+  try {
+    // TODO: replace with client.rest.
```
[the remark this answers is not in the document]
dsame: yes, it is a suggestion to unify the code style with the rest of repo. Some investigation of octakit needed - i do not know how to issue DELETE method currently

README.md
```diff
@@ -26,6 +26,16 @@ permissions:
 
 You can find more information about the required permissions under the corresponding options that you wish to use.
 
+## Statefulness
+
+If the action ends because of [operations-per-run](#operations-per-run) then the next run will start from the first
+unprocessed issue skipping the issues proceeded during the previous run(s). The state is reset when all the issues are
+proceeded. This should be considered for scheduling workflow runs.
```
msorens, line 33: Typos in the readme: 
s/proceeded/processed/g

16987814:
33077649_1753275318092946_6157577570121416704_n_1753275314759613

16987814:
118083165_3261329187287544_3724992751511924860_n_3261329180620878
120201425_3375869992500129_5334114439251769458_n_3375869985833463

PR #1075, opened 2023-08-11
BC: Do not run stale logic for newly stale objects

jsoref:
**Description:**

Simplify code for handling the transition from not-stale to stale. Do not run lots of code that makes no sense during this transition.

https://github.com/jsoref/stale-bot-debug/actions/runs/5828506506/job/15806335430#step:2:97

BEHAVIOR CHANGE:
This change will force action users to trigger two action runs if they want to both mark an item as stale and close it.

(You could technically write a workflow that called the action twice, it doesn't actually require two workflows.)

**Related issue:**

**Check list:**
- [x] Mark if documentation changes are required. -- _Possibly_. It depends on how good the documentation is. Based on how I encounter this action used in the real world (most projects are misconfigured and really don't understand what the actions does or how it works and certainly not how to achieve what they want), my bet is the documentation is sufficiently poor that it doesn't really, beyond a release note that this is a behavior change.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -288,7 +288,6 @@ class Issue {
         this.milestone = issue.milestone;
         this.assignees = issue.assignees || [];
         this.isStale = (0, is_labeled_1.isLabeled)(this, this.staleLabel);
-        this.markedStaleThisRun = false;
     }
     get isPullRequest() {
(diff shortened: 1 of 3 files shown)


dsame:
Hello @jsoref , thank you for the PR, it seems to be reasonable, we need some time to investigate it

Issue #1046, opened 2023-07-05
Only-label shoud act as a filter


ov7a:
**Description**/**Justification:**
I'm running a stale action with "only-labels" setting. I expect it to act as a filter, so only issues/prs labeled with particular labels are considered during the run. 

However, the current behavior is to do a full-scan. It drains the operation limit very fast.

dusan-trickovic:
Hello, @ov7a ! Thanks for creating the issue! We'll do some investigation and see what seems to be the problem :)

dsame:
Hello @ov7a , unfortunately it is not possible to filter out issues by label, but the problem of running out  API rate limit is expected to be solved with https://github.com/actions/stale/pull/1033 which is accessible now from main branch:
```
      - uses: actions/stale@v8
```

Did it help?

ov7a:
@dsame Thank you for the reply. Could you please provide more details on why it's not possible?

The statefulness is a great feature! However, I think it’s only a partial solution to the problem. Consider a case when a repository has 2000+ open issues, but only 100 of them have some label. Full scan is 10x times slower. It also wastes a lot resources.

dsame:
Hello @ov7a, to known what labels has the label we have to fetch the Issue first. This the limitation of GitHub REST API used by stale action. The migration to GraphQL is not planned at the moment. 

The fetching bunch of issue is just one GitHub API call, so it does not affect reaching limit of API rate greatly.

kurtqq:
@dsame can you also take a look at https://github.com/actions/stale/issues/837 ? it has some similarities

mraleph:
> Hello @ov7a, to know the labels of the issue, it has to be fetched first. This is the limitation of GitHub REST API used by stale action.

I don't think this is true. API has `labels` parameter to filter out issues based on the list of labels.

mraleph:
I made a PR: https://github.com/actions/stale/pull/1084

PR #1084, opened 2023-08-30
Filter issues by labels when listing if possible

mraleph:
**Description:**

* If `only-issue-labels` and `only-pr-labels` are not specified
  we limit the listing to `only-labels`.
* If `only-issue-labels` and `only-pr-labels` are specified
  but are the same, we limit the listing to these labels.
* If we don't need to make neither issues nor PRs as stale
  (`days-before-issue-stale` and `days-before-pr-stale` are
  negative) and stale label is the same for issues and PRs
  then we filter by this label.

This allows to significantly limit amount of processing done by this action on repos with large number of open issues.

**Related issue:**

https://github.com/actions/stale/issues/1046

**Check list:**

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -676,12 +676,33 @@ class IssuesProcessor {
     getIssues(page) {
         var _a;
         return __awaiter(this, void 0, void 0, function* () {
+            // Compute labels filter.
+            let labels = '';
+            // We can filter issues already when listing them if neither only-pr-labels
+            // nor only-issue-labels are specified or both are the same.
+            if (this.options.onlyPrLabels === '' &&
+                this.options.onlyIssueLabels === '') {
+                labels = this.options.onlyLabels;
+            }
+            else if (this.options.onlyPrLabels == this.options.onlyIssueLabels) {
+                labels = this.options.onlyIssueLabels;
+            }
+            // If we don't have to mark issues and pull requests stale and
+            // both the stale labels are the same then we can use it as a filter:
+            // because we only want to process stale issues.
+            if (!(0, should_mark_when_stale_1.shouldMarkWhenStale)(this._getDaysBeforeIssueStale()) &&
+                !(0, should_mark_when_stale_1.shouldMarkWhenStale)(this._getDaysBeforePrStale()) &&
+                this.options.stalePrLabel === this.options.staleIssueLabel &&
+                !labels.includes(this.options.staleIssueLabel)) {
+                labels += (labels !== '' ? ',' : '') + this.options.staleIssueLabel;
+            }
             try {
                 this.operations.consumeOperation();
                 const issueResult = yield this.client.rest.issues.listForRepo({
                     owner: github_1.context.repo.owner,
                     repo: github_1.context.repo.repo,
                     state: 'open',
+                    labels,
                     per_page: 100,
                     direction: this.options.ascending ? 'asc' : 'desc',
                     page
--- a/src/classes/issues-processor.ts
+++ b/src/classes/issues-processor.ts
@@ -562,12 +562,39 @@ export class IssuesProcessor {
 
   // grab issues from github in batches of 100
   async getIssues(page: number): Promise<Issue[]> {
+    // Compute labels filter.
+    let labels = '';
+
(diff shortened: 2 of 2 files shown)


dist/index.js
```diff
@@ -676,12 +676,33 @@ class IssuesProcessor {
...
+            let labels = '';
+            // We can filter issues already when listing them if neither only-pr-labels
+            // nor only-issue-labels are specified or both are the same.
+            if (this.options.onlyPrLabels === '' &&
+                this.options.onlyIssueLabels === '') {
+                labels = this.options.onlyLabels;
+            }
+            else if (this.options.onlyPrLabels == this.options.onlyIssueLabels) {
+                labels = this.options.onlyIssueLabels;
+            }
```
lrhn, line 689: (Microoptimizing for code size, this can be:
```js
  if (this.options.onlyPrLabels == this.options.onlyIssueLables) { 
    labels = this.options.onlyIssueLables || this.options.onlyLabels;
  }
```

Feel free to ignore.)

dist/index.js
```diff
@@ -676,12 +676,33 @@ class IssuesProcessor {
...
+            // If we don't have to mark issues and pull requests stale and
+            // both the stale labels are the same then we can use it as a filter:
+            // because we only want to process stale issues.
+            if (!(0, should_mark_when_stale_1.shouldMarkWhenStale)(this._getDaysBeforeIssueStale()) &&
+                !(0, should_mark_when_stale_1.shouldMarkWhenStale)(this._getDaysBeforePrStale()) &&
```
lrhn, line 694: Is the `(0, ...)` necessary?

I know the pattern can be used to call the function as a tear-off instead of a method, so not setting `this` during the call, but the same code pattern occurs in `issue-processor.ts` and doesn't use the `(0, ...)` hack there.

dist/index.js
```diff
@@ -676,12 +676,33 @@ class IssuesProcessor {
...
+            // nor only-issue-labels are specified or both are the same.
+            if (this.options.onlyPrLabels === '' &&
+                this.options.onlyIssueLabels === '') {
+                labels = this.options.onlyLabels;
+            }
+            else if (this.options.onlyPrLabels == this.options.onlyIssueLabels) {
+                labels = this.options.onlyIssueLabels;
+            }
+            // If we don't have to mark issues and pull requests stale and
+            // both the stale labels are the same then we can use it as a filter:
```
lrhn, line 691: The `:` at the end of the line should probably be a `,` or just removed.

I'd suggest a comma before "then". Maybe before "and" in the prior line. The long sentence is hard to read. (I'm easily confused, please help me! :grin:).

dist/index.js
```diff
@@ -676,12 +676,33 @@ class IssuesProcessor {
...
+            // because we only want to process stale issues.
+            if (!(0, should_mark_when_stale_1.shouldMarkWhenStale)(this._getDaysBeforeIssueStale()) &&
+                !(0, should_mark_when_stale_1.shouldMarkWhenStale)(this._getDaysBeforePrStale()) &&
+                this.options.stalePrLabel === this.options.staleIssueLabel &&
+                !labels.includes(this.options.staleIssueLabel)) {
```
lrhn, line 696: The `includes` doesn't require the text to be an entire label. If we have labels `stale` and `maybe-stale`, then the `includes('stale')` will match a `maybe-stale` label.
Consider over-engineering this as:
```js
!new RegExp('(?:^|,)' + this.options.staleIssueLabel + '(?:,|$)').test(labels)
```
If there can be surrounding whitespace around the label, then:
```js
!new RegExp('(?:^|,)\\s*' + this.options.staleIssueLabel + '\\s*(?:,|$)').test(labels)
````

(Or the, maybe, more memory-intensive, but less RegExp-y:
```js
  ![...labels.split(',').map(s=>s.trim())].includes(this.options.staleIssueLable)
```
)

lrhn (commented):
(Drive-by-comments. Sorry, not sorry :wink:)

dist/index.js
mraleph: This is generated code - does not make much sense to review it :) Also I don't think code size matters here.

PR #1110, opened 2023-11-24
Upgrade Node to v20

aparnajyothi-y:
**Description:**
Describe your changes.

**Related issue:**

**Check list:**


src/classes/issues-processor.ts
```diff
@@ -33,6 +33,13 @@ import {RateLimit} from './rate-limit';
 /***
  * Handle processing of issues for staleness/closure.
  */
+
+type state_reason_type =
+  | 'completed'
+  | 'reopened'
+  | 'not_planned'
+  | null
+  | undefined;
```
IvanZosimov, line 42: What was the reason for adding this type?

src/classes/issues-processor.ts
```diff
@@ -934,7 +941,7 @@ export class IssuesProcessor {
           repo: context.repo.repo,
           issue_number: issue.number,
           state: 'closed',
-          state_reason: this.options.closeIssueReason || undefined
+          state_reason: this.options.closeIssueReason as state_reason_type
```
IvanZosimov, line 944: In this code update, you changed the logic. The `this.options.closeIssueReason` is typed as a string.  The code `state_reason: this.options.closeIssueReason || undefined` means that if the `this.options.closeIssueReason` is an empty string we assign `state_reason` with undefined.
In your version of code this logic is changed.  If the `this.options.closeIssueReason` is an empty string you assign `state_reason` with an empty string. What is the reason for that?

src/classes/issues-processor.ts
aparnajyothi-y: While I was running npm run test few test and test suites are failed for the code

src/classes/issues-processor.ts
aparnajyothi-y: To resolve the failed tests for that particular line I have added the change and I missed the logic that I am assigning the empty string. Now I have updated the logic. Thank you for the clarification.

MaksimZhukov (approved):

marko-zivic-93 (approved):

Cabo999 (approved):

Cabo999 (approved):

PR #1152, opened 2024-04-10
Fix checking state cache (fix #1136), also switch to octokit methods

### LINKED ISSUE #1136: State restoration fails if a repo has many caches
**Description:**
If a repository has many caches being used between runs of `actions/stale`, such that the "_state" cache entry is not on the first page of results returned from the [list caches API](https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#list-github-actions-caches-for-a-repository), `actions/stale` will fail to restore its state and then will fail to update the state.

**Action version:**
9.0.0

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
1. Create a repo set up with `actions/stale` with a low `operations-per-run` and enough issues that a single run cannot process them all.
2. Have `actions/stale` run, which should create the "_state" cache entry.
3. Run other actions to create various other cache entries, until visiting `https://api.github.com/repos/{owner}/{repo}/actions/caches` no longer includes "_state" as it has been pushed to the second page of results.
4. Have `actions/stale` run again.

My test repo for this issue is at https://github.com/anomiex/test-stale.

**Expected behavior:**
For step 4, the state is restored correctly and processing continues from where the previous run left off.

**Actual behavior:**
In step 4, near the start, it reports
```
The saved state was not found, the process starts from the first issue.
```
Processing begins from the first issue. Then at the end it reports
```
Failed to save: Unable to reserve cache with key _state, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/master, Key: _state, Version: fa41d75081481069cfb6b92a5f83a94c6e06ef3ab2e6b762649ac5f86f46153f
```

**Analysis:**
The `checkIfCacheExists` function only checks the first page of results, using the default value of 30 entries per page.
https://github.com/actions/stale/blob/3f3b0175e8c66fb49b9a6d5a0cd1f8436d4c3ab6/src/classes/state/state-cache-storage.ts#L33-L46
While you could fix the bug by adding pagination so the relevant cache entry is found even if it's not on the first page, a more performant fix would be to make use of the `key` and `ref` parameters to the [list caches API](https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#list-github-actions-caches-for-a-repository) to specify the branch and key that we actually care about.

I suppose doing both would be an even better idea, just in case some repo is creating a lot of other cache entries with the "_state" prefix.

itchyny:
**Description:**
This PR fixes #1136, also fixes a TODO comment about using octokit methods.

**Related issue:**
- #1136

**Check list:**
  - These methods are not covered by existing tests, possibly because they interact with the GitHub APIs.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1618,9 +1618,13 @@ const getOctokitClient = () => {
 const checkIfCacheExists = (cacheKey) => __awaiter(void 0, void 0, void 0, function* () {
     const client = getOctokitClient();
     try {
-        const issueResult = yield client.request(`/repos/${github_1.context.repo.owner}/${github_1.context.repo.repo}/actions/caches`);
-        const caches = issueResult.data['actions_caches'] || [];
-        return Boolean(caches.find(cache => cache['key'] === cacheKey));
+        const cachesResult = yield client.rest.actions.getActionsCacheList({
+            owner: github_1.context.repo.owner,
+            repo: github_1.context.repo.repo,
+            key: cacheKey, // prefix matching
+        });
+        const caches = cachesResult.data['actions_caches'] || [];
+        return caches.some(cache => cache['key'] === cacheKey);
     }
     catch (error) {
         core.debug(`Error checking if cache exist: ${error.message}`);
@@ -1631,8 +1635,11 @@ const resetCacheWithOctokit = (cacheKey) => __awaiter(void 0, void 0, void 0, fu
     const client = getOctokitClient();
     core.debug(`remove cache "${cacheKey}"`);
     try {
-        // TODO: replace with client.rest.
-        yield client.request(`DELETE /repos/${github_1.context.repo.owner}/${github_1.context.repo.repo}/actions/caches?key=${cacheKey}`);
+        yield client.rest.actions.deleteActionsCacheByKey({
+            owner: github_1.context.repo.owner,
+            repo: github_1.context.repo.repo,
+            key: cacheKey,
+        });
     }
     catch (error) {
         if (error.status) {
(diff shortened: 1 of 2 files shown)


Hinaroos67:
> **Description:**
> This PR fixes #1136, also fixes a TODO comment about using octokit methods.
> 
> **Related issue:**
> - #1136
> 
> **Check list:**
> - [ ] Mark if documentation changes are required.
> - [ ] Mark if tests were added or updated to cover the changes.
>   - These methods are not covered by existing tests, possibly because they interact with the GitHub APIs.
>

Georgegriff:
I think i ran into this issue,  do you know if its likely this pr might progress?

itchyny:
I hope some maintainer will look into this PR, but maintenance activity of this action looks to be _stale_.

Dreamsorcerer:
@Jcambass @HarithaVattikuti Can someone merge this PR? The action is completely broken for many repos without this. We have stale issues that should have been closed weeks ago, but never close until this fix is released.

chiranjib-swain:
Hi @itchyny, thanks for the PR!
It looks like the formatting check failed during CI. The following file(s) need to be formatted with Prettier:
`src/classes/state/state-cache-storage.ts`
Could you please run Prettier locally and commit the updated formatting? You can do this with:
`npm run format`
This will ensure your changes pass the formatting check. Thanks!

chiranjib-swain (approved):

aparnajyothi-y (approved):

Issue #1137, opened 2024-02-22
Have one workflow ignore the state saved by another


rcomer:
**Description:**
Given two workflows in the same repo using this action, have one workflow ignore the state saved by another.  I think either of these two options would work for us:
* Make the cache key specific to the workflow
* Make statefulness optional with a config setting

**Justification:**
Over in [Matplotlib](https://github.com/matplotlib/matplotlib), we are using the action to slowly work through a large backlog of issues.  We want to keep notifications fairly minimal so have `operations-per-run` set low.  However, we don't want steps like removing the label after update or closing the issue to wait until we've worked through the whole backlog.  So when we updated to v9, we also added a second workflow using the action with `days-before-stale=-1` and a much higher `operations-per-run` (see my PR here https://github.com/matplotlib/matplotlib/pull/27523).  The two workflows run on alternate days.  Once #1136 is fixed, I believe what will happen is

- The original workflow processes a few issues and saves the state
- The second workflow finds that state, processes all the remaining issues and deletes the state
- The original workflow starts from the beginning

Since only the original workflow is adding the label, we still do not make progress through our backlog!  If each workflow has its own state, this would not be a problem.  Also if we could tell the second workflow not to look for a state, it would not be a problem.

**Are you willing to submit a PR?**

Willing, yes, but I do not have skills with typescript....

HarithaVattikuti:
Hello @rcomer 
Thank you for creating this feature request. We will investigate it and get back to you as soon as we have some feedback.

echang49:
Old issue but PR is created for this issue.

rcomer:
Thank you @echang49!

VOVELEE:
It will be really great if #1169 is merged. As mentioned in the PR I am hitting the same issue

PR #1134, opened 2024-02-02
github actions cache: Use more unique key

autoantwort:
**Description:**
If multiple jobs with stale steps exists in a repo we want a cache entry per run. 

**Related issue:**
https://github.com/actions/stale/issues/1133

Does not fully fixes the issue, since you still need different jobs instead of steps 

**Check list:**
- [ ] Mark if documentation changes are required.
- [ ] Mark if tests were added or updated to cover the changes.

PR #1169, opened 2024-08-02
Add cache-prefixing

echang49:
**Description:**
This changes allow the state saved to the global action cache to be stored with a prefix. Useful as `_state` is not a unique name and multiple workflows could save the cache as `_state`. Also useful as you currently cannot use this action twice for independent runs as the runs would pick up the same state. With this change, it is possible to have one run close all PRs with label `x` after 7 days and another run close all PRs except those with label `x` after 10 days.

Also updated the README.md to show proper permissions needed.

**Related issue:**
https://github.com/actions/stale/issues/1137

Please also look into merging https://github.com/actions/stale/pull/1152 as this isn't necessarily useful if caching isn't 100% fixed.

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ This can be achieved with the following [configuration in the action](https://do
 
 ```yaml
 permissions:
+  actions: write # for caching state
   contents: write # only for delete-branch option
   issues: write
   pull-requests: write
@@ -97,6 +98,7 @@ Every argument is optional.
 | [ignore-issue-updates](#ignore-issue-updates)                       | Override [ignore-updates](#ignore-updates) for issues only                  |                       |
 | [ignore-pr-updates](#ignore-pr-updates)                             | Override [ignore-updates](#ignore-updates) for PRs only                     |                       |
 | [include-only-assigned](#include-only-assigned)                     | Process only assigned issues                                                | `false`               |
+| [cache-prefix](#cache-prefix)                                       | Add a prefix to the stored cache                                            |                       |
 
 ### List of output options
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -547,6 +549,11 @@ If set to `true`, only the issues or the pull requests with an assignee will be
 
 Default value: `false`
 
+#### cache-prefix
+
+Beneficial so the action has a more unique cahce key. Useful when calling this action multiple times, independent of each other.
```
ccoVeille, line 554:
```suggestion
Beneficial so the action has a more unique cache key. Useful when calling this action multiple times, independent of each other.
```

README.md
```diff
@@ -547,6 +549,11 @@ If set to `true`, only the issues or the pull requests with an assignee will be
...
+An example for usage would be closing all PRs with `x` label after 7 days in one action and closing all PRs except for `x` label after 10 days in another.
```
ccoVeille, line 555:
```suggestion
An example for usage would be closing all PRs with `x` label after 7 days in one action, and closing all PRs except for `x` label after 10 days in another.
```

VOVELEE:
I am hitting similar issue but this time it is because I have couple of runners working on the same VM. In this situation the user (under which the first runner operates) creates the folder and file and remains owner preventing other runners (running under different context) to have access to the tmp folder.

autoantwort:
This solves the same problem as #1134. Here the user has to do manually work so solve this problem, with #1134 the user don't have to do anything. With this PR changes to the action configuration don't restart the process, with #1134 this will reset the progress.

echang49:
> This solves the same problem as #1134. Here the user has to do manually work so solve this problem, with #1134 the user don't have to do anything. With this PR changes to the action configuration don't restart the process, with #1134 this will reset the progress.

It's been a while so the full reasoning escapes me. I remember seeing your PR and the reasoning why I created a separate one was because:
* Not a fan of having hashed cache keys. Especially in the presence of other CI utilizing caches. One reason is that it becomes harder to determine where a cache is coming from. The use case that I had, a repository regularly exceeded the allowed cache limit. While some caches were ok to evict, others (like this one), should have been kept. With hashed cache keys, it becomes difficult to know which caches should be evicted or not.
* Seems like the possibility of having collisions with caching is still present if parameters are still present (if the params are all the same). I wanted the user to decide if they wanted to use the same cache or a different one. Didn't want to remove the choice from them.

Issue #1159, opened 2024-05-30
`actions: write` is too permissive


RedYetiDev:
## References
#1134 
<https://github.com/nodejs/help/pull/4396>

## Issue

Currently, the stale bot (and the runner in general) require `actions: write` permissions to modify existing caches but not to create new ones. If `actions: write` permissions aren't necessary to create caches, why are they needed to update them? Could a new `cache: write` permission be introduced for the runner?

HarithaVattikuti:
Hello @RedYetiDev 
Thank you for creating this issue. We will investigate it and get back to you as soon as we have some feedback.

ericLemanissier:
@RedYetiDev  you probably want to try my proposition from https://github.com/actions/stale/pull/1183.
For this, you can replace `uses: actions/stale@...`with `uses: actions/stale@ericLemanissier/patch-2` and remove `actions: write` from https://github.com/nodejs/help/blob/main/.github/workflows/close-stale-issues.yml

ericLemanissier:
@RedYetiDev  you probably want to try my proposition from https://github.com/actions/stale/pull/1237.
For this, you can replace `uses: actions/stale@...`with `uses: actions/stale@ericLemanissier/immutable_caches` and remove `actions: write` from https://github.com/nodejs/help/blob/main/.github/workflows/close-stale-issues.yml

Issue #1133, opened 2024-02-02
Error delete _state: [403] Resource not accessible by integration


autoantwort:
**Description:**
If a job has multiple `actions/stale` steps the creation of the cache fails.

**Action version:**
9

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
https://github.com/microsoft/vcpkg/blob/master/.github/workflows/stale.yml

**Expected behavior:**
No errors. 

**Actual behavior:**
Errors. See https://github.com/microsoft/vcpkg/actions/runs/7750221792

ericLemanissier:
Hello, the same error happens in here : https://github.com/conan-io/conan-center-index/actions/runs/7810903828/job/21305055330#step:2:1032

sophies927:
Hello, I am encountering the same error here: https://github.com/microsoft/onnxruntime/actions/runs/7917906948/job/21615103971#step:2:3615.

Pil0tXia:
Hi everyone, I'm encountering the same issue at https://github.com/apache/eventmesh/actions/runs/8675497930/job/23788986036 and I hope this log would be helpful to you.

Maybe I would revert back to v8 and see if this problem still exists.

shmishra99:
Granting `actions:write` to the stale workflow will allow it to manage issues cache as intended.

jeremystretch:
This appears to be the same issue described in #1131.

autoantwort:
Adding `actions: write` also fixes parts of the problem. The same cache entry is then still used across different cache steps

ericLemanissier:
Giving `actions: write` means the workflow has [all these permissions](https://docs.github.com/en/rest/authentication/permissions-required-for-fine-grained-personal-access-tokens?apiVersion=2022-11-28#repository-permissions-for-actions) and this is way too much, and not acceptable for a lot of projects.
https://github.com/actions/cache does not require these permissions, and still manages to save a cache across builds. Could the `stale` action be refactored to use the same trick as the `cache` action ? Or could it use the `cache` action under the hood ?

suyashgaonkar:
Hi @autoantwort , we did investigation with provided information for the issue. Tried to reproduce the issue with actions/stale v8 and v9 both, and everything seems to be working fine on our end. We did not get the mentioned error Warning: Error delete _state: [403] Resource not accessible by integration.
[...]

Issue #1131, opened 2024-01-29
Stale workflow fails to override cache.


mshima:
**Description:**
We have a stale workflow: https://github.com/jhipster/generator-jhipster/blob/main/.github/workflows/stale.yml

At first run it successfully generates a cache entry.

After that, the cache is never updated and is reused, so cached entries are never analyzed again.
See builds 599-605 at https://github.com/jhipster/generator-jhipster/actions/workflows/stale.yml 

See 2 consecutive days have same cached issue numbers:
https://github.com/jhipster/generator-jhipster/actions/runs/7689412024/job/20951825772
https://github.com/jhipster/generator-jhipster/actions/runs/7682085625/job/20936000418

The error reported is:
Warning: Error delete _state: [403] Resource not accessible by integration

Are we missing a configuration? A permission maybe? There is no info in the docs.

**Action version:**
v9

**Platform:**
- [ ] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
Links above.

**Expected behavior:**
A new cache should be written at every run.

**Actual behavior:**
New cache are not been overridden

HarithaVattikuti:
Hello @mshima
Thank you for creating this issue. We will investigate it and get back to you as soon as we have some feedback.

csuermann:
Looks like issues already processed in previous runs never get closed due to the cache.

_.github/workflows/stale.yaml_:
```YAML
name: 'Close stale issues'
on:
  workflow_dispatch:
  schedule:
    - cron: '30 1 * * *'
permissions:
  issues: write

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v9
        with:
          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
          days-before-stale: 30
          days-before-close: 5
```

```
Run actions/stale@v9
Cache Size: ~0 MB (307 B)
/usr/bin/tar -xf /home/runner/work/_temp/f4844f75-67b0-4546-8244-42b90d2f9d54/cache.tzst -P -C /home/runner/work/xxx --use-compress-program unzstd
Cache restored successfully
state: restored with info about 12 issue(s)
Starting the stale action process...
Processing the batch of issues  #1  containing  11  issues...
[#239] Issue #239
[#238] Issue #238
[#237] Issue #237
[#235] Issue #235
[...]

PR #1183, opened 2024-10-16
do not attempt to delete existing cache

ericLemanissier:
**Description:**
I don't think it's necessary to delete existing cache before overwriting it. https://github.com/actions/cache does not seem to do it. The problem with deleting a cache entry is that it requires to give `actions: write` permissions, which means the workflow has [all these permissions](https://docs.github.com/en/rest/authentication/permissions-required-for-fine-grained-personal-access-tokens?apiVersion=2022-11-28#repository-permissions-for-actions) and this is way too much, and not acceptable for a lot of projects.

Also, do not check if cache exists during restore. It introduces a TOCTOU issue, and is not needed because `restoreCache` returns `undefined` if the cache does not exist, cf https://github.com/actions/toolkit/tree/main/packages/cache

**Related issue:**
fixes https://github.com/actions/stale/issues/1159
fixes https://github.com/actions/stale/issues/1133
fixes https://github.com/actions/stale/issues/1131

**Check list:**

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1592,8 +1592,6 @@ const fs_1 = __importDefault(__nccwpck_require__(7147));
 const path_1 = __importDefault(__nccwpck_require__(1017));
 const os_1 = __importDefault(__nccwpck_require__(2037));
 const core = __importStar(__nccwpck_require__(2186));
-const github_1 = __nccwpck_require__(5438);
-const plugin_retry_1 = __nccwpck_require__(6298);
 const cache = __importStar(__nccwpck_require__(7799));
 const CACHE_KEY = '_state';
 const STATE_FILE = 'state.txt';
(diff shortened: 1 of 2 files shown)


ericLemanissier:
CI results available in https://github.com/ericLemanissier/stale/pull/21/checks

ericLemanissier:
@suyashgaonkar  what do you think of this change ?

ericLemanissier:
@aparnajyothi-y @suyashgaonkar what to you think of this change ?

ericLemanissier:
well, it looks like caches are actually immutable.
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow indicates `Use caching when you want to reuse files that don't change often between jobs or workflow runs`, so it seems to me that github cache is not the proper solution for storage of stale state, because it changes at each run.
Why not use `actions/upload-artifact` to save the state, then 
```
- uses: actions/download-artifact@v4
  with:
    run-id: ${{ github.run_id -1 }}
```

Issue #1156, opened 2024-04-27
Actions fails if rate Limit is not enabled


ggabijaa:
**Description:**
When running action on schedule errors due to rateLimit. The actions is marked as successful but logs were showing errors:
<img width="550" alt="image" src="https://github.com/actions/stale/assets/59805033/50fc8e93-1b30-432d-b43c-802fa73bd150">

Is this expected behavior? Should action error out when rateLimit, not enabled?

**Action version:** V9

**Platform:** Rocky Linux

**Runner type:** Self-hosted

**Workfkow config:**  
```
name: Mark Stale Pull Requests

on:
  schedule:
    - cron: '0 1 * * *'

permissions:
  issues: write
  pull-requests: write


jobs:
  stale:
    runs-on: [self-hosted, rocky]
    steps:
      - uses: actions/stale@v9
        with:
          days-before-stale: 3
          stale-pr-label: 'stale'
          labels-to-remove-when-unstale: 'stale'
          days-before-close: -1
```

HarithaVattikuti:
Hello @ggabijaa 
Thank you for creating this issue. We will investigate it and get back to you as soon as we have some feedback.

martin-majlis-s1:
When I check the documentation - https://docs.github.com/en/enterprise-server@3.13/admin/configuring-settings/configuring-user-applications-for-your-enterprise/configuring-rate-limits#about-rate-limits-for-github-actions - I can see, that it says:

> By default, the rate limit for GitHub Actions is disabled. Because GitHub Enterprise Server can handle temporary spikes in usage without performance degradation, this rate limit is intended to protect against sustained high load. We recommend leaving the rate limit disabled unless you are experiencing performance problems. In some cases, GitHub Support may recommend that you enable a rate limit for GitHub Actions.

Not having a rate limit is expected state and it should not be logged as an error - https://github.com/actions/stale/blob/main/src/classes/issues-processor.ts#L648

What do you think?

nunosousa-dev:
Is this still being analyzed? At least a flag to not check for rate limites could be considered, or just remove this error print for GH Enterprise instances. Thank you.

suyashgaonkar:
Hi @ggabijaa  , We tried to investigate the issue with the provided details on a self hosted runner and we see that the workflow seems to work fine without any rate limit issue. It will really helpful if you could share a link to a public  repository to help us with further investigation on the issue.

suyashgaonkar:
[...]

PR #1185, opened 2024-10-18
fix: warning if rate limit is disabled

nunosousa-dev:
**Description:**
This will add a warning if rate limit is disabled (for GitHub Enterprises use cases).
Today we got an error message when running our workflow, we should instead get a warning since this is a feature from GitHub Enterprise and not an error. 

I'm making it as a warning because it's best practices to have it enabled so we could at least inform users that's disabled.

**Related issue:**
#1156 

**Check list:**

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -744,10 +744,30 @@ class IssuesProcessor {
             const logger = new logger_1.Logger();
             try {
                 const rateLimitResult = yield this.client.rest.rateLimit.get();
-                return new rate_limit_1.RateLimit(rateLimitResult.data.rate);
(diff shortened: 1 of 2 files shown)


chiranjib-swain:
Thank you  for raising and drafting a solution for this issue . We appreciate your initiative and effort in contributing to the project.


A comprehensive fix has now been merged into `main` via [PR #1300](https://github.com/actions/stale/pull/1300), which resolves the issue discussed here. If you encounter any related problems after updating to the latest version, please let us know.


With this, we'll mark your pull request as resolved and proceed to close it.  
Thanks again for your valuable contribution!

PR #1219, opened 2025-03-07
Allow option to exclude specific weekdays from total time

### LINKED ISSUE #564: Allow option to exclude weekends from total time
## The problem
We would like to keep the total time open pretty low - but weekends cause items to be flagged early. For example we would like it to be 3 weekdays - but currently a PR opened on Friday would be stale on Monday.

## The solution
Allow an option to exclude weekends from the total time open.

alpaca-tc:
**Description:**

Add support for excluding specific weekdays when calculating elapsed days for stale issues and PRs. This is particularly useful for organizations that want to consider only business days.

- Add new `exclude-weekdays` option to specify which days to exclude (0-6, where 0 is Sunday)
- Implement weekday exclusion logic in elapsed days calculation
- Add tests to verify business days calculation

```yaml
- uses: actions/stale@v9
  with:
    days-before-stale: 5
    days-before-close: 2
    exclude-weekdays: '0,6'  # Exclude weekends
```

**Related issue:**

Fixed #564

**Check list:**

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -375,6 +375,7 @@ const is_valid_date_1 = __nccwpck_require__(891);
 const is_boolean_1 = __nccwpck_require__(8236);
 const is_labeled_1 = __nccwpck_require__(6792);
 const clean_label_1 = __nccwpck_require__(7752);
+const elapsed_millis_excluding_days_1 = __nccwpck_require__(4101);
 const should_mark_when_stale_1 = __nccwpck_require__(2461);
 const words_to_list_1 = __nccwpck_require__(1883);
 const assignees_1 = __nccwpck_require__(7236);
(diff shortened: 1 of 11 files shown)


dist/index.js
```diff
@@ -2333,6 +2334,62 @@ function isValidDate(date) {
...
+        if (excludeWeekdays.includes(from.getDay())) {
+            elapsedMillis -= startOfNextDayFrom.getTime() - from.getTime();
+        }
+        if (excludeWeekdays.includes(to.getDay())) {
+            elapsedMillis -= to.getTime() - startOfDayTo.getTime();
+        }
+        const excludeWeekdaysCount = countWeekdaysBetweenDates(startOfNextDayFrom, startOfDayTo, excludeWeekdays);
+        elapsedMillis -= excludeWeekdaysCount * DAY;
+    }
+    return elapsedMillis;
```
alpaca-tc, line 2388: If you don't use options, the existing implementation is utilized, so even if there are bugs, the impact is minimal. A simple O(N) calculation logic is used. While an O(1) logic could also be possible, it would be a bit more complex.

PR #1248, opened 2025-05-07
Update README.md

GhadimiR:
**Description:**
To maintain state, this action uses GitHub Actions Cache, via the API. One of the operations it needs to perform in the event of a cache conflict is to [delete the existing cache](https://github.com/actions/stale/blob/816d9db1aba399a7f70277f1a2b01a4d21497fdd/src/classes/state/state-cache-storage.ts#L39), and this [requires the `actions: write` permission](https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-github-actions-caches-for-a-repository-using-a-cache-key).

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ This can be achieved with the following [configuration in the action](https://do
 
 ```yaml
 permissions:
+  actions: write
   contents: write # only for delete-branch option
   issues: write
   pull-requests: write


ericLemanissier:
Alternatively, https://github.com/actions/stale/pull/1237 could be considered, avoiding unnecessary priviledge elevation

GhadimiR:
Thanks for weighing in @ericLemanissier, I've had a look over your PR and while I agree that there's some merit to the idea of using unique cache keys to just generate new caches instead of resolving a conflict, that isn't the kind of thing we'd do lightly. Would prefer to make this small change to fix the docs and consider something like that for a future major release.

suyashgaonkar (approved):

HarithaVattikuti (approved):

Issue #1293, opened 2025-10-03
Support `exempt-issue-types` option


veryfancy:
**Description:**

https://github.com/actions/stale/issues/1187 added support for a `only-issue-types` option.

Other similar options like `only-issue-labels` have blocklist counterparts, like `exempt-issue-labels`. We can do the same and give `only-issue-types` a counterpart `exempt-issue-types`.

**Justification:**

My team wanted to use this to prevent certain types of issues from being closed as stale. However, we have a lot of untyped issues, and configuring our workflow using an `only-issue-labels` allowlist instead of a blocklist looks like it will prevent untyped issues from being closed as stale. That's not the behavior we'd prefer.

**Are you willing to submit a PR?**

Yes, I have provided a PR.

aparnajyothi-y:
Hello @veryfancy, Thank you for creating this issue and we will get back to you with some feedback :)

v-chiranjib-swain:
This has been addressed by #1294, which adds support for the `exempt-issue-types` option. The PR has been merged into main, and the option will be available once a new version of _actions/stale_ is released.

Thanks for the suggestion and for contributing the fix! 🎉

Issue #1187, opened 2024-11-03
Filter based on issue type


Bibo-Joshi:
**Description:**
Introduce a parameter `only-issue-types` that ensures that only issues of a specific [type](https://docs.github.com/en/issues/tracking-your-work-with-issues/configuring-issues/managing-issue-types-in-an-organization) are set to stale and closed.

**Justification:**
With the introduction of types, GitHub as given organizations a new issue attribute that can be used for categorization/classification of issues. For different issue types it would be desirable to have different stale settings just like we currently can have for labels. E.g. in my organization, we'll be using the new issue type for separating questions from feature requests. Questions should go stale faster.

**Are you willing to submit a PR?**
I'm open to try so, but will need some pointers, especially on testing. I'm not familia with TypeScript (or very familiar with JS at all …)

mahabaleshwars:
Hello @Bibo-Joshi,
Thank you for creating this issue. We will investigate it and provide feedback as soon as we have some updates.

JosephTLyons:
zed.dev is switching over to using issue types as well to categorize features and bugs - would love to have this configuration option.

maxwxyz:
This is very much needed, otherwise we have duplicates for labels and types now.

JosephTLyons:
One thing to note is that GitHub's API doesn't even support filtering by type, as far as I can tell. So until that lands, getting that feature in the bot is totally blocked.

https://github.com/orgs/community/discussions/140874#discussioncomment-11582660

JasonStoltz:
Looks like they've added Issue Type to the Reset API.

Mister-Hope:
Any progress with this one? I think there are no blocking things with it.

andig:
Stale bot is currently unusable without support for issue types...

Bibo-Joshi:
@mahabaleshwars I'd still be open to giving this a shot if you could provide me with a few pointers on how to get started :)

Tobi-mmt:
@Bibo-Joshi Since it is pretty much the same as the `exempt-issue-labels` option right now, I guess its something like:
1. check out the repo and read the https://github.com/actions/stale/blob/main/CONTRIBUTING.md
2. extend the options enum like here https://github.com/actions/stale/blob/f78de9780efb7a789cf4745957fa3374cbb94fd5/src/enums/option.ts#L15
[...]

PR #1255, opened 2025-05-19
Add `only-issue-types` option to filter issues by type

Bibo-Joshi:
**Description:**
Tries to implement a filter based on the issue type on issues as described in #1187. I tried following the suggestions in https://github.com/actions/stale/issues/1187#issuecomment-2875989020. Copilot did help me, though I reviewed everything manually and made sure to follow the instructions in the contribution guide. Hopefully I didn't miss anything …

**Related issue:**
closes #1187

**Check list:**
- [x] Mark if documentation changes are required.
- [x] Mark if tests were added or updated to cover the changes.

PR #1294, opened 2025-10-03
Add `exempt-issue-types` option for blocklist filter by issue type

veryfancy:
**Description:**

Introduce a parameter `exempt-issue-types` that ensures that issues from a list of types will not be marked stale or closed. This option applies only to issues; pull requests are always ignored by the issue-type filter.

This is a blocklist counterpart to the allowlist option `only-issue-types` added in:

- https://github.com/actions/stale/issues/1187

**Related issue:**

- https://github.com/actions/stale/issues/1293

**Check list:**


maxwxyz:
@luketomlinson this is similar to #1255

Bibo-Joshi (changes requested):
Hi. As a heads-up, I see that `main.ts` was not updated, which will probably lead to the input being ignored. See #1298, were I discovered that for `only-issue-types` :)

Bibo-Joshi (approved):
Sweet :)

veryfancy:
Bump to this. 🙏

veryfancy:
Bump to this. 🙏

laureanobrs:
Bump. Currently, there’s no way to process untyped issues together with only certain typed issues. This feature would make that possible. :pray:

TNATALI:
Bump this ya'll. Literally we are waiting on this PR to be merged and available. Would do a ton of good for the community.

veryfancy:
Fixed conflict with dist.js and updated behavior to keep it from filtering pull requests (to match the fix from https://github.com/actions/stale/pull/1338).

v-chiranjib-swain:
@veryfancy , thanks for this update. I tested the branch locally and with real Issue Types, including its interaction with `only-issue-types`, and it works as expected.

Could we document that `exempt-issue-types` takes precedence when both lists contain the same type and add a regression test for that case? This would make the behavior clear and prevent future regressions.

veryfancy:
@v-chiranjib-swain Thanks for testing this. I documented the precedence and added a regression test covering a type present in both lists.

v-chiranjib-swain (approved):

v-priya-kinthali (approved):

v-HarithaVattikuti (approved):

PR #1177, opened 2024-09-11
Update commenter type

jayudey-wf:
**Description:**
I was noticing that several of the places where I had instrumented the stale action it seemed like some of the pr's were having the label's appropriately applied but the action was never coming back around and closing them after the determined stale period.

I ran a small version of [this](https://github.com/actions/stale/blob/main/src/classes/issues-processor.ts#L551) as part of a small script and saw that the output indicated that the github actions comment was a 'Bot' type.

**Related issue:**
Add link to the related issue.

**Check list:**
- [x] Mark if documentation changes are required. (n/a)
- [x] Mark if tests were added or updated to cover the changes.

PR #1309, opened 2025-12-04
don't account for bot comments

jayudey-wf:
**Description:**
I was noticing that several of the places where I had instrumented the stale action it seemed like some of the pr's were having the label's appropriately applied but the action was never coming back around and closing them after the determined stale period.

I ran a small version of [this](https://github.com/actions/stale/blob/main/src/classes/issues-processor.ts#L551) as part of a small script and saw that the output indicated that the github actions comment was a 'Bot' type.

I had previously created this change set here, https://github.com/actions/stale/pull/1177 but given all the changes i thought to just start a new.

**Related issue:**

**Check list:**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ $ npm install
 Build the typescript and package it for distribution.
 
 ```bash
-$ npm run build && npm run pack
+$ npm run build
 ```
 
 Run the tests :heavy_check_mark:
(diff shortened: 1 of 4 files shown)


AdamGS:
just also ran into this isse, I have [PR](https://github.com/vortex-data/vortex/pull/5304) that also keep flip-flopping.
Screenshot 2026-03-11 at 11 17 37

Issue #1184, opened 2024-10-17
Last updated date appears to be incorrect


michaelmior:
**Description:**
As the excerpt from the log below shows, I have a pull request that is identified as having been updated within the past several days even though the last update I can see is nearly one year ago.

```
[#514] Pull request #514
  [#514] Found this pull request last updated at: 2024-10-15T01:53:46Z
```

**Action version:**
v9

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
I am unsure how to reproduce this issue with a fresh repository, but [here is an example](https://github.com/apache/calcite/actions/runs/11377902079/job/31652795554) of a run of this action that exhibits the problem.
For example, look at [pull request 512](https://github.com/apache/calcite/pull/514).
It has not been updated in more than a year, but the action seems to think it was updated within the past several days.

**Expected behavior:**
The correct date of the last update should be identified.

**Actual behavior:**
The date of last update is not identified correctly.

michaelmior:
I do see the `updated_at` date returned by the API is within the past several days, so I assume this is the root of the issue.

HarithaVattikuti:
Hello @michaelmior 👋,
Thank you for reporting this issue. We will investigate it and get back to you as soon as we have some feedback.

dktapps:
Yep, running into this with a bunch of stale PRs on my repo too. For some reason a bunch of them got unmarked as stale 3 days ago.

Adam262Alloy:
I am noticing this in our repos as well. Taking three example PRs for example, they each have a comment from the Stale Github action that they were quickly marked and unmarked as stale on 1/6/25. Neither PR has user activity on that day (they are automated dependabot type PRs), and all should be stale in that they are older than our configured threshhold for staleness.

![image](https://github.com/user-attachments/assets/9285ab50-2bed-4d90-ac09-d058c4210bf8)

When I look in the Github action logs, I can see the action processed these PRs consecutively in the same batch, and they have very similar updated at messages:

PR 1 --  `Found this pull request last updated at: 2025-01-06T22:22:13Z`
PR 2 --  `Found this pull request last updated at: 2025-01-06T22:22:13Z`
PR 3 --  `Found this pull request last updated at: 2025-01-06T22:22:14Z`

wandbjake:
[...]

PR #1310, opened 2025-12-26
Fix: Increase tolerance in isDateMoreRecentThan to 30s

shamoon:
**Description:**

This increases the "tolerance" of checking `updated_at` vs. `markedStaleOn` from 15 to 30 seconds. See linked thread but for us recently stalebot has been marking issues stale and then un-marking them because of its own actions, seemingly just because the timestamps are > 15s:

> ```
>   [#11353] Found this pull request last updated at: 2025-12-23T03:32:45Z
>   ...
>   [#11353] Pull request marked stale on: 2025-12-23T03:32:17Z
>   [#11353] Checking for comments on pull request since: 2025-12-23T03:32:17Z
>   ...
>   [#11353] Pull request has been updated since it was marked stale: true
>   [#11353] Pull request has been commented on: false
>   [#11353] Remove the stale label since the pull request has been updated and the workflow should remove the stale label when updated
> ```

Of course I do not know exactly what the ideal tolerance would be but 30 seconds seems relatively as "safe" as 15. Nor do I know the underlying reason the tolerance is no longer sufficient, perhaps there has been drift in the GH API over time. 

Of course alternatively this could be made configurable, but Im not sure how widespread this really is to warrant that.

**Related issue:**
See #1184 (it would close this in my case but Im not 100% certain if OP there has the same exact cause).

**See also:**
#816

**Check list:**
- [ ] ~~Mark if documentation changes are required.~~
- [ ] ~~Mark if tests were added or updated to cover the changes.~~

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -803,9 +803,9 @@ class IssuesProcessor {
                 issueLogger.info(`marked stale this run, so don't check for updates`);
                 yield this._removeLabelsOnStatusTransition(issue, labelsToRemoveWhenStale, option_1.Option.LabelsToRemoveWhenStale);
             }
-            // The issue.updated_at and markedStaleOn are not always exactly in sync (they can be off by a second or 2)
-            // isDateMoreRecentThan makes sure they are not the same date within a certain tolerance (15 seconds in this case)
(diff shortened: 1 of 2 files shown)


shamoon:
Happy to close this in favor of #1311 depending on wishes of the maintainers. Sorry for all the noise!

PR #1311, opened 2025-12-26
Enhancement: ignore stale labeling events

shamoon:
**Description:**
This PR is an alternative to #1310 that I believe will solve the issue more gracefully. The idea is if `issue.updated_at` is more recent than `markedStaleOn` _and_ the issue has updated since stale (preserving the 15s 'tolerance' that seems to have mostly worked for a long time) we then check for and "ignore" the adding the stale label event(s).

In other words, dont count adding the stale label towards un-stale-ing an issue. Hopefully preventing things like:

Screenshot 2025-12-25 at 4 11 13 PM

Other notes:

- I updated the PR since opening it to use cached events that are already retrieved in `getLabelCreationDate`, so _no new API calls_
- I did use ChatGPT for generating the tests, mostly building the mock client stuff, the rest is mine entirely

```sh
Statistics:
Processed PRs: 1
Test Suites: 29 passed, 29 total
Tests:       1359 passed, 1359 total
Snapshots:   0 total
Time:        5.499 s, estimated 7 s
```

**Related issue:**
See https://github.com/actions/stale/issues/1184 but this is a broader scope

**Check list:**
- [ ] ~~Mark if documentation changes are required.~~ (I dont think so, but happy to if needed)

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -417,7 +417,6 @@ class IssuesProcessor {
         this.addedLabelIssues = [];
         this.addedCloseCommentIssues = [];
         this._logger = new logger_1.Logger();
-        this._issueEventsCache = new Map();
         this.options = options;
         this.state = state;
         this.client = (0, github_1.getOctokit)(this.options.repoToken, undefined, plugin_retry_1.retry);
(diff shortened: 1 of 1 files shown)


shamoon:
What the @copilot comments about the API calls does not take into account is:

- The new logic in this PR only happens if `updated_at` is after `markedStaleOn` by at least 15 seconds (the "tolerance" that seems to have been working for a long time) and of course if `remove-stale-when-updated` is `true` **and** `!issue.markedStaleThisRun`
- If the above fails (e.g. 300 events isnt far enough back, which I assume is not that common) the behavior is essentially unchanged from what it does now.

mala-sap (approved):
👍

src/classes/issues-processor.ts
```diff
@@ -77,6 +77,7 @@ export class IssuesProcessor {
   readonly addedCloseCommentIssues: Issue[] = [];
   readonly statistics: Statistics | undefined;
   private readonly _logger: Logger = new Logger();
+  private readonly _issueEventsCache: Map<number, IIssueEvent[]> = new Map();
```
chiranjib-swain: Hi @shamoon ,
I noticed the `_issueEventsCache` is used, but given that each issue is processed only once per workflow run and `getLabelCreationDate` is only called once per issue, the cache never actually results in a cache hit. Since the state doesn’t persist across runs, it adds complexity without benefit.

I recommend removing the cache to simplify the code and avoid confusion. Let me know if there are plans to use it differently—happy to discuss!

Lexxi5:
I want to unsubsidized

Issue #1333, opened 2026-04-10
Ignore issues which has linked open PRs


mmzeynalli:
**Description:**
Sometimes one collects many PRs before releasing a new version of the library. Most of these PRs are linked to issues and will be closed when the merge happens. Till that time, even if there is no activity on the issue, work is going on PR, and the issue logically should stay active.

**Justification:**
Holding a bunch of PRs (linked to issues) ready for the next version publish.

**Are you willing to submit a PR?**

Would love to, but this is not a priority for me right now.

lmvysakh:
Hello @mmzeynalli ,
Thank you for creating this issue. We will investigate it and get back to you as soon as we have some feedback.

PR #1357, opened 2026-08-04
feat: add exempt-issues-with-open-linked-pr option

somaz94:
**Description:**

Adds an opt-in `exempt-issues-with-open-linked-pr` option. When enabled, an issue is left alone while an open pull request is linked to it in a way that will close it once merged. Work is still happening on the pull request, so the issue is not really inactive.

Three notes on the design, since this touches the operations budget:

- **It costs an operation, so it runs last.** Following the note above the draft-PR check in `issues-processor.ts` ("this check is so far below because it cost one read operation"), the lookup only runs after every cheaper check has passed, so it is limited to issues that would otherwise be marked stale.
- **Disabled by default,** so existing workflows keep the exact same `operations-per-run` budget. There is a test asserting the lookup is never called when the option is off.
- **Only linked pull requests count, not mentions.** It uses the GraphQL `closedByPullRequestsReferences(includeClosedPrs: false)` field, which is the "Development" link a `Closes #123` creates. The REST timeline was the other option, but its `cross-referenced` events also fire on a plain mention, which would exempt issues that merely got named in some pull request. GraphQL is reached through the existing `getOctokit` client, so there is no new dependency.

If the query fails, the error is logged and the issue is processed as usual rather than being silently exempted.

**Related issue:**

Related to #1333.

**Check list:**

Validation, run locally on node 24:

- `npm run format-check`, `npm run lint`, `npm run build` all pass, and `dist/` rebuilds byte-identically
- `npm test` goes from 29 suites / 1360 tests to 30 / 1366, no regressions
- I checked the new tests actually fail without the feature: disabling the exemption breaks exactly the "should not stale the issue when an open pull request will close it" case, and removing the pull-request guard breaks exactly the case covering that a pull request cannot have a linked pull request


somaz94:
This is out of draft now and ready for review.

The five checks on `c54266d` — `Basic validation`, `Check dist/`, `e2e tests`, `Licensed`, `Code scanning` — are all sitting at `action_required`, i.e. the first-run approval gate for an outside contributor, so none of them have actually executed yet. Could a maintainer kick off the run?

To save you the risk of spending that approval on a red run, I re-validated locally today. The branch is already current with `main` (no rebase needed), and:

- `npm test` — 30/30 suites, 1366/1366 tests pass
- `npm run format-check` and `npm run lint` — clean
- `npm run build` — reproduces the committed `dist/index.js` byte-for-byte

Happy to rebase or adjust anything if you'd rather see the option shaped differently.

